PR #20, opened 2023-10-31
feat(component): support json reference

donch1989:
Because

- we want to render the json reference on the fly.

This commit

- support json reference


donch1989:
related to https://github.com/instill-ai/community/issues/465

PR #54, opened 2024-02-05
feat: add tasks to component definition

jvallesm:
Because

- New fields have been added to the `ConnectorDefinition` and `OperatorDefinition` entities. Some of them will be read from the definition JSON schema but tasks require some manipulation.

This commit

- Extract task name, title and definition from task schemas.
- Tests also check for other newly added fields that should be present in the definition JSON schemas.

# Notes

I started by rendering the source URL here as well. I didn't love it because it couples the code with a package that isn't even being imported. I don't love, either, writing this in the definition JSON schema. There will be a lot of repetition and schemas are complex enough to read as to have braindead code.

```go
baseSourceURLPath := "github.com/instill-ai/operator/blob/main/pkg/"
def.SourceUrl = baseSourceURLPath + strings.ReplaceAll(def.Id, "-", "")
```

However, this won't work when we have versioned components and will complicate the expression. Plus, some connectors (`gcp`, `airbyte`, `instill-model`) don't match the ID to their path. I think the best solution for now is having the URL in the JSON schema and, in the future, lint or generate automatically the field to make sure it matches the path.


donch1989 (approved):
LGTM

PR #56, opened 2024-02-14
fix: fix `instillFormat` validation when using `semi-structured` and `structured data`

donch1989:
Because

- The `instillFormat` validation has a bug when dealing with `semi-structured` and `structured` data. In such cases, base64 encoding should not be checked.

This commit

- Skips `instillFormat` base64 validation when `instillFormat` is classified as `semi-structured` or `structured`.

Note

- The current approach is somewhat hard-coded. We should optimize the flow in the future.

### DIFF
--- a/pkg/base/formats.go
+++ b/pkg/base/formats.go
@@ -28,15 +28,23 @@ type InstillAcceptFormatsSchema []string
 
 func (s InstillAcceptFormatsSchema) Validate(ctx jsonschema.ValidationContext, v interface{}) error {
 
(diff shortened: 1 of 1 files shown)


donch1989:
Hi @jvallesm ,
Please review this PR. It should resolve the [issue](https://github.com/instill-ai/operator/pull/61#issuecomment-1943255785). Thanks!

jvallesm (approved):
🙇 thanks for addressing this!

PR #61, opened 2024-02-26
chore: rename blockchain connector to Application connector 

jvallesm:
Because

- https://github.com/instill-ai/protobufs/pull/274 removed the Blockchain
  connector type

This commit

- Updates `compogen` with the new connector type


jvallesm:
Merging as it only modifies `compogen`, production code isn't affected.

PR #64, opened 2024-03-06
feat: simplify `openapi_specifications` to `data_specifications`

donch1989:
Because

- Related to this https://github.com/instill-ai/protobufs/pull/282, currently, we have an `openapi_specification` field in the connector/operator definition. The original purpose was that we could generate an OpenAPI schema and have some UI to let the user debug on an individual component. However, it turns out that we don't really need that OpenAPI schema, and it makes our response very large and hard to read. 

This commit

- Simplifies `openapi_specification` into `data_specification` by preserving only the input and output JSON schema and removing the nested structure of the OpenAPI spec.


pkg/base/connector.go
```diff
@@ -27,9 +27,9 @@ type IConnector interface {
 	// Add definition
 	AddConnectorDefinition(def *pipelinePB.ConnectorDefinition) error
 	// Get the connector definition by definition uid
-	GetConnectorDefinitionByUID(defUID uuid.UUID, resourceConfig *structpb.Struct, componentConfig *structpb.Struct) (*pipelinePB.ConnectorDefinition, error)
+	GetConnectorDefinitionByUID(defUID uuid.UUID, resourceConfig *structpb.Struct, component *pipelinePB.ConnectorComponent) (*pipelinePB.ConnectorDefinition, error)
```
donch1989, line 30: Note:
The reason we want to pass component data into the definition is that some fields are dynamically generated. For example, in the Instill Model Connector, the enum inside the definition data JSON schema is generated by the connector setting on the fly. So we need to pass component data into it and generate a correct JSON schema.

The current implementation design looks a little bit weird, but it works. I think we can improve the structure in the future.

jvallesm: 👍 Thanks for the clarification, I wondered why we had these unused fields but with this info I realised that each connector implementation (e.g. `restapi.Init()` must implement this interface and there we make use of these fields.

We'll need to update now the `connector` and `operator` repos with the new interface, right?

jvallesm (commented):
Looks good!

PR #79, opened 2024-04-02
refactor: make the component definitions JSON to contain a single definition per component

donch1989:
Because

- Originally, we could have multiple definitions in a component (e.g., Airbyte connector). However, we have now made all components have a single definition. Therefore, there is no need to use an array of definitions in the definition JSON files.

This commit

- Makes the component definitions JSON to contain a single definition per component.


jvallesm (approved):
😌 🧹

PR #80, opened 2024-04-03
chore: update generate Readme

iamnamananand996:
Because

- update generate Readme

This commit

- update generate Readme

### DIFF
--- a/pkg/connector/googlecloudstorage/v0/README.mdx
+++ b/pkg/connector/googlecloudstorage/v0/README.mdx
@@ -1,0 +1,39 @@
+---
+title: "Google Cloud Storage"
+lang: "en-US"
+draft: false
+description: "Learn about how to set up a VDP Google Cloud Storage operator https://github.com/instill-ai/instill-core"
+---
+
+The Google Cloud Storage component is an operator that allows users to upload data to Google's Cloud Storage.
+It can carry out the following tasks:
+
+- [Upload](#upload)
+
+## Release Stage
+
+`Alpha`
+
+## Configuration
+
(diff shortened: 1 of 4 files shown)


jvallesm:
> or make gen-doc, which was lost when the operator repo was merged

I'll add a commit with the Makefile if you're okay with that @iamnamananand996

pkg/connector/googlecloudstorage/v0/README.mdx
jvallesm, line 1: ℹ️ This has removed your added files but as soon as you add the build command they will come back

iamnamananand996: Hi @jvallesm, after executing the build command, I am not able to generate the file.

jvallesm: You should only add the build tag to the components whose README files you were originally adding, i.e.:

```
pkg/connector/archetypeai/v0/main.go
pkg/connector/googlecloudstorage/v0/main.go
pkg/connector/googlesearch/v0/main.go
pkg/connector/pinecone/v0/main.go
```

You added the build tag to _every_ component and some of them are not supported, so the `make` command will fail.

pkg/operator/start/v0/main.go
jvallesm, line 1: `start` and `end` are special components that shouldn't be documented as other components / operators, they don't follow the same structure and frontend expectations. Please don't tag these 2.

They should be moved away from the `operator` space after we introduced the iterator component.

jvallesm (approved):

PR #99, opened 2024-04-23
feat(compogen): use jsonref when generating the README

donch1989:
Because

- The `compogen` tool didn't use `jsonref` to render the definition, some component readme cannot be generated.

This commit

- Uses `jsonref` when generating the README.
- Renames `resource_specification` to `connection_specification`.
- Removes the validation for an empty `connection_specification`.

Note

- We still do not support nested structures yet.


pkg/connector/bigquery/v0/README.mdx
jvallesm, line 1: For automatic generation, we can add[ a `generate` tag in the component's `main` file,](https://github.com/instill-ai/component/blob/main/pkg/connector/archetypeai/v0/main.go#L1) then running `make gen-doc` will generate all the README.mdx files.

tools/compogen/pkg/gen/schema.go
```diff
@@ -14,6 +14,6 @@ type property struct {
 }
 
 type objectSchema struct {
-	Properties map[string]property `json:"properties" validate:"gt=0,dive"`
+	Properties map[string]property `json:"properties" validate:"dive"`
```
jvallesm, line 17: I'd rather avoid empty objects in the definitions and keep this validation but if that's not possible I'm fine with loosening the restrictions. Did you encounter any other case with empty object schemas apart from `resource_specification`?

tools/compogen/pkg/gen/readme.go
```diff
@@ -255,12 +283,20 @@ func parseResourceProperties(o *objectSchema) []resourceProperty {
...
-		// We can safely access the order pointer because it has been
-		// previously validated by the caller.
-		props[*prop.Order] = prop
+	idx := 0
+	for k := range propMap {
+		props[idx] = propMap[k]
+		idx += 1
 	}
 
+	// Note: The order might not be consecutive numbers.
```
jvallesm, line 292: :ah: I was wondering about these changes

tools/compogen/pkg/gen/readme.go
```diff
@@ -255,12 +283,20 @@ func parseResourceProperties(o *objectSchema) []resourceProperty {
...
 
 	props := make([]resourceProperty, len(o.Properties))
-	for _, prop := range propMap {
-		// We can safely access the order pointer because it has been
-		// previously validated by the caller.
-		props[*prop.Order] = prop
+	idx := 0
+	for k := range propMap {
+		props[idx] = propMap[k]
+		idx += 1
```
jvallesm, line 289:
```suggestion
	for k := range propMap {
		props[k] = propMap[k]
```

Or, alternatively

```go
	for k, prop := range propMap {
		props[k] = prop
```

jvallesm (changes requested):
Thanks for taking care of this 💛 

A couple of changes are required:
- The tests aren't passing with these changes, they need to be updated
  - It'd be good to add a GHA for this package using `make test`
- We need to add the generation tags to the components that can have autogenerated documentation. That way, the integration action Naman is working on will take them into account.

If you prefer it, I can have a look at this items and contribute to this branch.

pkg/connector/bigquery/v0/README.mdx
donch1989: oh.. I didn't notice we have the tag here. Let me add them

tools/compogen/pkg/gen/schema.go
donch1989: Yeah.. Looks like we can add the validation back. It still works.

tools/compogen/pkg/gen/readme.go
donch1989: `props` is an array with a size of `len(o.properties)`, and the `order` might not be in the range `0:len(o.properties)`. Therefore, I changed the implementation to use sorting.

tools/compogen/pkg/gen/readme.go
donch1989: We don't require the order of the properties to be consecutive from `0:len`. They might not be unique or consecutive.

tools/compogen/pkg/gen/schema.go
donch1989: Let me do a quick try

tools/compogen/pkg/gen/readme.go
jvallesm: Yeah when I implemented this I didn't think `order` didn't need to be consecutive or in range, so 👍 to using sorting here. My suggestion here is that using `idx` is unnecessary. `k` already starts at 0 and is incremented in each loop iteration.

jvallesm: argh 🤦 right, sorry for the noise

tools/compogen/pkg/gen/schema.go
donch1989: The Instill Model connector will encounter some issues with this (I guess it is caused by `oneof`). However, we are also discussing removing the "external mode" for the Instill Model connector. The problem will be resolved if we decide to remove it.

jvallesm (approved):

PR #100, opened 2024-04-24
feat: test compogen on PR and push events

jvallesm:
Because

- `compogen` tests are broken

This commit

- fixes tests
- adds action to check `compogen` when changed


jvallesm:
@donch1989 please consider these changes for #99 . 1007dc8b3d52ef25ef2cd1f40402937776f67720 is only needed in case we can't remove empty objects.

PR #104, opened 2024-04-25
chore: reintroduce nonempty object validation

jvallesm:
Because

- We want to avoid empty objects in our definitions if possible.
- New **Connection** section is empty for some connectors.

This commit

- **chore: reintroduce nonempty object validation**
- **fix: remove connection section if empty**


donch1989:
Hi @jvallesm 
This PR has some problems when putting it in the pipeline backend. I'll add a commit to make it work.

PR #110, opened 2024-05-07
feat: allow global API key on OpenAI connector

jvallesm:
Because

- We need connectors like OpenAI to use globally configured secrets to run executions without credentials.

This commit

- Injects a global secret for the OpenAI API key on intialization.
- Transforms a keyword value in that configuration parameter into the secret.

### DIFF
--- a/pkg/connector/openai/v0/main.go
+++ b/pkg/connector/openai/v0/main.go
@@ -6,6 +6,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/gabriel-vasile/mimetype"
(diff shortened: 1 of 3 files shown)


pkg/connector/openai/v0/main.go
```diff
@@ -35,39 +39,93 @@ var (
...
 
-func (c *connector) CreateExecution(sysVars map[string]any, connection *structpb.Struct, task string) (*base.ExecutionWrapper, error) {
+// The connection parameter is defined with snake_case, but the
+// environment variable configuration loader replaces underscores by dots,
+// so we can't use the parameter key directly.
+func readFromSecrets(key string, s map[string]any) string {
+	sanitized := strings.ReplaceAll(key, "_", "")
+	if v, ok := s[sanitized].(string); ok {
+		return v
+	}
```
jvallesm, line 83: ❓ If we used `camelCase` in connection params this could be simplified. But it requires a migration so I didn't take that path. Still interesting to consider, since we've recently discussed unifying cases in definition files.

donch1989: Yes, I think we'll use kebab-case for all component fields soon. We can remove this workaround after we change all the fields to kebab-case.

PR #113, opened 2024-05-08
feat: Chunhao/ins 4298 slack component dev

chuang8511:
Because

- we want more new component in vdp.

This commit

- first stage of slack component

Note
- The unit test has not been done.


pkg/connector/slack/v0/config/definition.json
```diff
@@ -0,0 +1,46 @@
...
+  "icon": "assets/xxx.svg",
+  "icon_url": "",
+  "id": "slack",
+  "public": true,
+  "spec": {
+    "connection_specification": {
+      "$schema": "http://json-schema.org/draft-07/schema#",
+      "additionalProperties": false,
+      "properties": {
+        "api_key": {
```
donch1989, line 17: Please try to align the naming with the vendor. Slack is using `token`, so let's use `token` here.

chuang8511:
@donch1989 
Do we need this for slack?
It seems we do not need this, so I do not implement.

```
// main.go
func (c connector) Test(sysVars map[string]any, connection *structpb.Struct) error {

	return nil
}


// connector_test.go (refer to other test code)
func TestConnector_Test(t *testing.T) {
```

chuang8511:
Note
- test passed locally, will check why the test failed here.
- Will need to resolve the confliction.
  - [ref](https://instill-ai.slack.com/archives/C016F66FX9Q/p1715249935803859)

chuang8511:
Will comment back the test later and check why it failed in action but not in local.

chuang8511:
Commit to be signed

PR #125, opened 2024-05-14
fix: Fix the bug of setting channelID

chuang8511:
Because

- channelID is wrong to set

This commit

- fix set channelID bug

### DIFF
--- a/pkg/connector/slack/v0/apiFunctions.go
+++ b/pkg/connector/slack/v0/apiFunctions.go
@@ -20,7 +20,7 @@ func loopChannelListAPI(e *execution, isPublic bool, channelName string) (string
 			return "", err
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/connector/slack/v0/apiFunctions.go
```diff
@@ -20,7 +20,7 @@ func loopChannelListAPI(e *execution, isPublic bool, channelName string) (string
 			return "", err
 		}
 
-		targetChannelID := getChannelID(channelName, slackChannels)
+		targetChannelID = getChannelID(channelName, slackChannels)
 
 		if targetChannelID != "" {
 			break
```
donch1989, line 26: you can return here directly

chuang8511: @donch1989 Modified

donch1989 (approved):

PR #126, opened 2024-05-14
fix(slack): add `instillSecret: true` to `token` field

chuang8511:
Because

- We want the token is secret.

This commit

- make the token secret

### DIFF
--- a/pkg/connector/slack/v0/config/definition.json
+++ b/pkg/connector/slack/v0/config/definition.json
@@ -22,7 +22,6 @@
             "instillAcceptFormats": [
               "string"
             ],
-            "instillCredentialField": false,
             "instillSecret": true,
             "instillUIOrder": 0,
             "title": "token",
@@ -30,6 +29,7 @@
           }
         },
         "required": ["token"],
+        "instillEditOnNodeFields": ["token"],
         "title": "Slack Connection",
         "type": "object"
       }
--- a/pkg/connector/slack/v0/config/definition.json
+++ b/pkg/connector/slack/v0/config/definition.json
@@ -23,12 +23,13 @@
               "string"
             ],
             "instillCredentialField": false,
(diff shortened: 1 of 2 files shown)


pkg/connector/slack/v0/config/definition.json
```diff
@@ -23,12 +23,13 @@
               "string"
             ],
             "instillCredentialField": false,
```
donch1989, line 25: Please remove this line.

pkg/connector/slack/v0/config/definition.json
```diff
@@ -23,12 +23,13 @@
...
+            "instillSecret": true,
             "instillUIOrder": 0,
             "title": "token",
             "type": "string"
           }
         },
-        "required": [],
+        "required": ["token"],
         "title": "Slack Connection",
```
donch1989, line 33: Add additional `"instillEditOnNodeFields": ["token"]` to make it appears on the Console node.

donch1989 (changes requested):

donch1989:
Found another bug: if the conversations array length is 0, it will return nil, but we should return an empty array. You can try make the start_date larger then now.

donch1989:
Please add `"instillUIMultiline": true` to `message` in TASK_WRITE_MESSAGE

chuang8511:
> Please add `"instillUIMultiline": true` to `message` in TASK_WRITE_MESSAGE
It is originally true in TASK_WRITE_MESSAGE

image

donch1989:
> > Please add `"instillUIMultiline": true` to `message` in TASK_WRITE_MESSAGE
> 
> @donch1989 It is originally true in TASK_WRITE_MESSAGE
> 
> image

Oh... Okay. My eyes are blurred.

donch1989 (approved):

PR #129, opened 2024-05-15
chore(openai): update Instill Credit supported model list

donch1989:
Because

- We'll support Instill Credit for text_to_image, text_to_speech, and speech_recognition tasks.

This commit

- Updates the list of Instill Credit supported models.


donch1989:
Hi @jvallesm 
Please check this PR. Currently, we need to add this map to the JSON schema so that Console can provide hints.

jvallesm (approved):
Oh right, I'd forgotten about this! Thanks for noticing, good catch!

PR #132, opened 2024-05-17
ci(Makefile): add test target

pinglin:
Because

- local unit testing is essential while running `go test ./...` currently requires a `-tag ocr` due to the [docconv](https://github.com/sajari/docconv) package using [tesseract](https://tesseract-ocr.github.io)

This commit

- add `make test`


donch1989:
The command doesn't work on my side. I'll check it later.

donch1989 (approved):

PR #138, opened 2024-05-20
feat: add pdf component

chuang8511:
Because

- We want users to transform PDF to markdown format file

This commit

- Transform PDF to markdown in VDP.

Note
- Python code should have its own test code. So, I mock `cmd`.
  - However, the current pdfTransformer is PoC for the first phase, so I did not add the test code for python.


pkg/operator/pdf/v0/main.go
```diff
@@ -0,0 +1,89 @@
...
+		switch e.Task {
+		case taskConvertPdfToMarkdown:
+			inputStruct := ConvertPdfToMarkdownInput{}
+			err := base.ConvertFromStructpb(input, &inputStruct)
+			if err != nil {
+				return nil, err
+			}
+
+			scriptPath := "/component/pkg/operator/pdf/v0/python/pdfTransformer.py"
+			pythonInterpreter := "/opt/venv/bin/python"
```
jvallesm, line 72: Consider having these as constants

pkg/operator/pdf/v0/main.go
```diff
@@ -0,0 +1,89 @@
...
+			cmd := exec.Command(pythonInterpreter, scriptPath)
+
+			outputStruct, err := convertPdfToMarkdown(inputStruct, cmd)
```
jvallesm, line 75: Conceptually, the command variables and executor initialisation belong in the `convert` method. `Execute` should only be responsible of transforming from / to structpb and multiplexing the logic to each task. Initialising the command outside of the method doesn't have any benefit here.

```suggestion
			outputStruct, err := convertPdfToMarkdown(inputStruct)
```

pkg/operator/pdf/v0/convertPdfToMarkdown_test.go
```diff
@@ -0,0 +1,51 @@
...
+		filepath string
+	}{
+		{
+			name:     "Convert pdf file",
+			filepath: "testdata/test.pdf",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			b, err := os.ReadFile(test.filepath)
```
jvallesm, line 21: If you're not going to check the output of the conversion, you can just pass some random bytes here

pkg/operator/pdf/v0/python/pdfTransformer.py
jvallesm, line 1: I find executing custom Python code from our main Go executor a bit complex.

- Did you look for Go PDF->Markdown converters that we can execute as a command or import as a package?
- Did you consider other Python packages that can be executed directly as a command with arguments, e.g. `https://github.com/VikParuchuri/marker`?

> Python code should have its own test code. So, I mock cmd. 

If `pdfplumber` is our best shot, I'd definitely unit-test our custom Python code. [testscript](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript) might be helpful here (it's used for testing the `compogen` command in this repo) in Go tests I think it's much more valuable to add Python to the testing environment and running the execution without mocks, checking the output of the converted PDF.

pkg/operator/pdf/v0/config/definition.json
```diff
@@ -0,0 +1,18 @@
+{
+  "available_tasks": [
+    "TASK_CONVERT_PDF_TO_MARKDOWN"
+  ],
+  "custom": false,
+  "documentation_url": "https://www.instill.tech/docs/latest/vdp/operators/pdf",
+  "icon": "assets/pdf.svg",
+  "icon_url": "",
```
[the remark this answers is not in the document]
chuang8511: > it isn't part of the protobuf contract anymore.
I will drop it.
But, I want to know what it means in this context.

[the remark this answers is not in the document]
jvallesm: note: it is the way `console` accessed the icon before the `icon` field was introduced. The asset was added in `console` and `icon_url` referenced its public URL.

We inverted the logic with `icon`, making `component` the source of truth for the component details and implementation. Now `console` scrapes this repo during build, adding the needed assets for each component.

pkg/operator/pdf/v0/mockCommand.go
```diff
@@ -0,0 +1,43 @@
+package pdf
+
+import (
+	"encoding/json"
+	"io"
+)
+
+type MockCommandRunner struct {
```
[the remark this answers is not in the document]
chuang8511: Thanks for this advice! It is useful!
As a previous Ruby developer, I am not so familiar with those tools.
I have searched this solution for a while.. I am not sure why I did not find it.
Thank you again!!

chuang8511 (commented):
I replied comments and will push the modification soon!

chuang8511:
image
I was trying to look for where we set this Pdf.
But, it seems it transforms in frontend.

pkg/mock/generator.go
```diff
@@ -1,3 +1,6 @@
 package mock
 
 //go:generate minimock -g -i github.com/instill-ai/component/pkg/base.UsageHandler -o ./ -s "_mock.gen.go"
+//go:generate minimock -g -i ./../operator/pdf/v0.commandRunner -o ./ -s "_mock.gen.go"
```
chuang8511, line 4: @jvallesm 
Before we have the component remotely, how do you set the minimock last time?
It seems that I have to add this back after we release the code.
Is the understanding correct?

chuang8511: Except this part, others are ok. So, I request review first!

pkg/operator/pdf/v0/convertPdfToMarkdown_test.go
```diff
@@ -0,0 +1,51 @@
...
+		name     string
+		filepath string
+	}{
+		{
+			name:     "Convert pdf file",
+			filepath: "testdata/test.pdf",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
jvallesm: Did you have a look at this? Are you going to adopt it or should we resolve this thread?

jvallesm:
> image I was trying to look for where we set this Pdf. But, it seems it transforms in frontend.

I see, I think it would be beneficial to have a `title` field in the base task object, so we only rely on the frontend transformation as a fallback. In cases where we have acronyms like here we can't just use the simplistic transformation (`_` -> ` ` + Title case). Please report this to frontend / product so we can prioritise this. For now I don't think this should prevent us from shipping the feature.

jvallesm (commented):
Thanks for addressing my suggestions. There are a few unaddressed comments but they're design / cosmetic suggestions rather than blocking change requests. Let me know what you think about them, if you intend to address any of them, and let's move forward when all the conversations are resolved.

pkg/operator/pdf/v0/main.go
chuang8511: Thanks for your feedback!

> However, if possible we should avoid modifying the production code just because it's more convenient for testing*.

I also agree on that! It is also a hard decision for me.
In Ruby, we can easily mock command without dependency injection. 
But, I am not quite sure if it is possible to do that. 
If we can do it in Go, we would love to do it. 
But, I searched for a while, I could not find the proper solution.

> You just need the pdfplumber dependencies in whatever environment you're running go test.

It indeed concerns me.
I treat Python code as a kind of external service, which is similar to API.
So, for me, it is the similar concept to mock API here.

pkg/mock/generator.go
```diff
@@ -1,3 +1,6 @@
...
+//go:generate minimock -g -i io.WriteCloser -o ./ -s "_mock.gen.go"
+
```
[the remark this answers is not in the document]
chuang8511: @jvallesm 
Do you mean we need to remove the blank line?
If yes, I met this error if I do it.
`pkg/mock/generator.go:5: error reading pkg/mock/generator.go: unexpected EOF`

Do you happen to know about it?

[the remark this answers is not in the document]
jvallesm: I meant removing the line, yes. I actually detected it because I ran `go generate` with the modification I suggested in the same file, so it should work. 🤔 I don't know what can be happening on your env but I don't feel too strongly about this so you can keep the blank line if you don't find a quick solution.

pkg/mock/usage_handler_mock.gen.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by http://github.com/gojuno/minimock (dev). DO NOT EDIT.
+// Code generated by http://github.com/gojuno/minimock (v3.3.9). DO NOT EDIT.
```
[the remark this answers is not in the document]
chuang8511: @jvallesm 
I tried to fix my version for a while.
But, I am not quite sure why I cannot install 3.3.6.
Do you happen to know this issue?

```
chunhao@MacBook-Pro-8 component % go get github.com/gojuno/minimock/v3
go: github.com/gojuno/minimock/v3@v3.3.9 requires go >= 1.22; switching to go1.22.3
go: upgraded go 1.21 => 1.22
go: added toolchain go1.22.3
go: upgraded github.com/gojuno/minimock/v3 v3.3.6 => v3.3.9
chunhao@MacBook-Pro-8 component % go get github.com/gojuno/minimock/v3@3.3.6
go: github.com/gojuno/minimock/v3@3.3.6: invalid version: unknown revision 3.3.6
```

pkg/operator/pdf/v0/convert_pdf_to_markdown_test.go
```diff
@@ -0,0 +1,62 @@
...
+			// Metadata: map[string]string{
+			// 	"title": "Test",
+			// },
+		}
+		mockOutputBytes, err := json.Marshal(mockOutput)
+		c.Assert(err, qt.IsNil)
+
+		mockRunner.CombinedOutputMock.Expect().Return(mockOutputBytes, nil)
+
+		output, err := convertPDFToMarkdown(input, mockRunner)
```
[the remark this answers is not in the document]
chuang8511: >  I'd rather have warnings for low coverage or missing documentation and add fewer but higher quality tests and comments. What do you think about this?

I agree on that, and I also care about the coverage.
But, even we input an easy binary, it will be hard to increase the test coverage eventually.

The essence of this low test coverage is the Python code is still too complicate.
And, I do not think we should not take too much time on unit test for PoC stage.
There are still several logic that we will need to revamp in the future.
I'd love to add more test cases when we move to the next stage.
For now, I think Python code as an external service in the perspective of Go repo.

chuang8511:
@jvallesm 
https://github.com/instill-ai/component/pull/138#discussion_r1611114093

I am not quite familiar with GitHub UXUI.
I have taken this out and fixed all Pdf to PDF already!
Just sync this information here with you.

pkg/operator/pdf/v0/convert_pdf_to_markdown_test.go
```diff
@@ -0,0 +1,62 @@
...
+	test := struct {
+		name     string
+		filepath string
+	}{
+		name:     "Convert PDF file",
+		filepath: "testdata/test.pdf",
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		c := qt.New(t)
```
jvallesm, line 24: Sorry, about [this last comment](https://github.com/instill-ai/component/pull/138#issuecomment-2126706733), I hadn't realised you had adopted quicktest. I saw `t.Run` and thought the tests still relied on `testing` only.

```suggestion
	c := qt.New(t)

	test := struct {
		name     string
		filepath string
	}{
		name:     "Convert PDF file",
		filepath: "testdata/test.pdf",
	}

	c.Run(test.name, func(c *qt.C) {
```

pkg/operator/pdf/v0/main.go
jvallesm: >  But, I am not quite sure if it is possible to do that. If we can do it in Go, we would love to do it.

I'm not sure this can be done cleanly, I wouldn't spend too much effort there.

About mocking the command vs running it in the tests, I understand your point in treating it as an API, I'd approach it differently but it isn't a blocker on my side.

pkg/mock/usage_handler_mock.gen.go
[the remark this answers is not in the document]
jvallesm: try `go install github.com/gojuno/minimock/v3/cmd/minimock@v3.3.6`

pkg/operator/pdf/v0/convert_pdf_to_markdown_test.go
chuang8511: @jvallesm 
I will modify it!

But, I read the [document](https://pkg.go.dev/github.com/frankban/quicktest#section-readme).
image

It seems the function for testing.Run and quicktest.Run are similar.
And, even we use `t.Run`, we still can use quicktest for assertion.
Do you know the different between them? Why currently the codebase choose `c.Run` rather than `t.Run`?

Thank you in advance!

jvallesm: > It seems the function for testing.Run and quicktest.Run are similar.
> And, even we use t.Run, we still can use quicktest for assertion.
> Do you know the different between them?

[`quicktest.C` is a wrapper over `testing.TB`](https://github.com/frankban/quicktest/blob/master/quicktest.go#L75) so many functions are interchangeable. The reason for defining `c` at the base of your test is saving you the creation of `c` in each subtest (a couple of lines here, more as your test function grows). It also allows you to run, for example, `c.Parallel`. But in this case is consistency about the tooling and conciseness.

> Why currently the codebase choose c.Run rather than t.Run

I proposed quicktest because I find its assertion style useful and readable. I've been using it for years and I'm happy with it so, since no one was against it, I kept using it out of convenience.

jvallesm (dismissed):
🏋️ 💦 Thanks for the thorough revision.

This is good to go 🚀

chuang8511:
For the reviewers @jvallesm (( cc'd @donch1989 
I tested with Dockerfile, and it seems the code is compiled.
So, it seems I cannot [execute the code with the way by the path.](https://github.com/instill-ai/component/pull/138/files#diff-79f88919f8ac3e50fd71fe42f0bafce6b27182bbd0e5192bb448adc141978ef1R69) 

So, I am testing to put all python code in Go file and execute it directly, which I also do not like.
But, it seems to be the easy way to deal with it.

I will push my code on Friday and notify you again to sync with you the result.

donch1989:
@chuang8511 
You can try to use [go embed](https://pkg.go.dev/embed). This can help you embed non-golang resource in the golang package.

donch1989:
@chuang8511
I'll merge this PR first. We can fix any bugs later if there are any.

donch1989 (approved):

PR #139, opened 2024-05-24
feat: add task chunk text

chuang8511:
Because

- we want to enrich the functions to chunk text for knowledge base

This commit

- add the task to chunk


external/langchaingo/textsplitter/token_splitter.go
```diff
@@ -0,0 +1,82 @@
+package textsplitter
+
+import (
+	"fmt"
+
+	"github.com/pkoukk/tiktoken-go"
```
jvallesm, line 6: 👀 In `pipeline-backend-cloud` we use https://github.com/hupe1980/go-tiktoken, please consider using a single package across our codebase.

Since this is going to be used by an operator, I'd avoid making external calls so, if we _need to_ keep this package, I'd use [this loader](https://github.com/pkoukk/tiktoken-go-loader).

external/langchaingo/textsplitter/token_splitter.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+const (
+	// nolint:gosec
```
jvallesm, line 10: When adding linter exceptions, please document why they're needed.

external/langchaingo/textsplitter/token_splitter.go
```diff
@@ -0,0 +1,82 @@
...
+	var tk *tiktoken.Tiktoken
+	var err error
+	if s.EncodingName != "" {
+		tk, err = tiktoken.GetEncoding(s.EncodingName)
+	} else {
+		tk, err = tiktoken.EncodingForModel(s.ModelName)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("tiktoken.GetEncoding: %w", err)
+	}
```
jvallesm, line 57: 🌷 `tokenizer` can be an unexported field in `TokenSplitter`, it only depends on the splitter options so you can initialise it on `NewTokenSplitter`.

external/langchaingo/textsplitter/token_splitter.go
```diff
@@ -0,0 +1,82 @@
...
+type TokenSplitter struct {
+	ChunkSize         int
+	ChunkOverlap      int
+	ModelName         string
+	EncodingName      string
+	AllowedSpecial    []string
+	DisallowedSpecial []string
+}
+
+func NewTokenSplitter(opts ...Option) TokenSplitter {
```
jvallesm, line 27: The value of this struct is in its methods so I'd return a pointer. There's quite a lot of literature about pointers vs values so the discusion is nuanced. I like the rationale in this [post](https://preslav.me/2024/04/23/two-reasons-to-prefer-struct-pointers-in-golang/)

```suggestion
func NewTokenSplitter(opts ...Option) *TokenSplitter {
```

external/langchaingo/textsplitter/token_splitter.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (s TokenSplitter) splitText(text string, tk *tiktoken.Tiktoken) []string {
+	splits := make([]string, 0)
+	inputIDs := tk.Encode(text, s.AllowedSpecial, s.DisallowedSpecial)
+
+	startIdx := 0
+	curIdx := len(inputIDs)
+	if startIdx+s.ChunkSize < curIdx {
+		curIdx = startIdx + s.ChunkSize
+	}
```
jvallesm, line 71:
```suggestion
	if s.ChunkSize < curIdx {
		curIdx = s.ChunkSize
	}
```

external/langchaingo/schema/chat_message_history.go
jvallesm, line 1: There are some files where everything is commented out. We should not include them.

jvallesm (changes requested):
❓ Is the code in `external/langchaingo` fully copied from the `langchaingo` project? Why not importing the package? If not, what are the main differences? The project is actively maintained so by duplicating its code we're losing any update they make (e.g. a vulnerability fix). I was reviewing this code, then realised it comes from somewhere else.

operator/text/v0/chunk_text.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if s.Separators == nil {
+		s.Separators = []string{"\n\n", "\n", " ", ""}
+	}
+}
+
+func chunkText(input ChunkTextInput) (ChunkTextOutput, error) {
+	var split textsplitter.TextSplitter
+	setting := input.Strategy.Setting
+	setting.SetDefault()
```
jvallesm, line 75: What's the point in having settings as user input (e.g. allowed special) if they're going to ve overriden? Shouldn't the `default` field in the config schema determine the default behaviour, rather than the Go code?

operator/text/v0/chunk_text.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	if setting.ChunkMethod == "Token" {
+		tkm, err := tiktoken.EncodingForModel(setting.ModelName)
+		if err != nil {
+			return ChunkTextOutput{}, err
+		}
+		token := tkm.Encode(input.Text, setting.AllowedSpecial, setting.DisallowedSpecial)
+		output.TokenCount = len(token)
+	}
```
jvallesm, line 124: This can go in the `switch` statement above

jvallesm: As mentioned in a previous comment, I'd avoid external calls in this document. You can set an offline BPE loader by using https://github.com/pkoukk/tiktoken-go-loader

jvallesm (commented):
> 1. About test case
> There are several functions that we should mock to avoid uncertainty of unit test. However, it will damage the code structure. So, I decide not to mock `split.SplitText` & `tkm.Encode`

Why would these methods be undeterministic? Given the same model / encoding and input, text should be split or tokenised consistently.

> 2. I tried to refactor the part of `switch e.Task`. However, it damaged the readability when I kept trying to abstract the code.

Could you expand on what you had in mind for this refactorisation and what you tried to achieve? This `switch` pattern was added not so long ago to avoid code duplication so I'd like to know if it's confusing or missing anything.

> 3. I want to keep the Go version in this feature branch. So, I commented out the parts that are required to upgrade Go version when I cherry-pick the code.

Not in the scope of this PR but I'm for updating Go to the latest version. The current version is already preventing us from upgrading certain packages. cc @donch1989

external/langchaingo/textsplitter/token_splitter.go
chuang8511: @jvallesm 
> Why not importing the package? If not, what are the main differences? 

There is no difference. I chose this way because of this discussion.
- [discussion](https://linear.app/instill-ai/issue/INS-4499/[component][text-operator]-add-task-chunk-text-for-more-advanced#comment-1da16232)
In short, the main reason is that importing all package needs to upgrade Go version.
And, I am not quite sure the risk about updating Go version.
So, I want to keep updating Go version in another branch.

> The project is actively maintained so by duplicating its code we're losing any update they make (e.g. a vulnerability fix).

I agree! I am thinking about we transfer to import the package directly after we upgrade Go version.
However, I am not quite sure about this topic. 
Do you have experience on this part?
- [It has its own data model and interfaces for connecting LLMs, database I/O, etc., which is a bit too heavy for a single Instill Component. We probably want to implement the code ourselves instead of importing it as a whole Go package.](https://linear.app/instill-ai/issue/INS-4499/[component][text-operator]-add-task-chunk-text-for-more-advanced#comment-fd95adca)

chuang8511: I checked Task_Split_By_Token, and it used `"github.com/pkoukk/tiktoken-go"`
- [code](https://github.com/instill-ai/component/blob/main/operator/text/v0/split.go#L4)

Is it different between `pipeline-backend-cloud` & `pipeline-backend` now?
Should we modify `pipeline-backend-cloud`?

I am not quite sure about whether [it is a good source](https://github.com/hupe1980/go-tiktoken).
It seems the star / fork count is low. 
Is it a standard to judge?

operator/text/v0/chunk_text.go
chuang8511: > Shouldn't the default field in the config schema determine the default behaviour, rather than the Go code?

I also think so. 
However, there is an unknown bug that I could not fix. 
I also feel sad about it. 😢 
- [Ref](https://instill-ai.slack.com/archives/C016F66FX9Q/p1716909862447969?thread_ts=1716560732.412659&cid=C016F66FX9Q)

But, in a positive way, it decreases the risk of the bugs come from other places.

operator/text/v0/chunk_text.go
chuang8511: [Replied here](https://github.com/instill-ai/component/pull/139#discussion_r1618440907)

chuang8511:
> 1. About test case
> There are several functions that we should mock to avoid uncertainty of unit test. However, it will damage the code structure. So, I decide not to mock split.SplitText & tkm.Encode
> Why would these methods be undeterministic? Given the same model / encoding and input, text should be split or tokenised consistently.

split functions belong to external services. If there is an update in external service, the result could be failure.
In my opinion, unit test should mock the external service to focus on what we are testing.
And, the logic of the external service should contain their own test.


> I tried to refactor the part of switch e.Task. However, it damaged the readability when I kept trying to abstract the code.
> Could you expand on what you had in mind for this refactorisation and what you tried to achieve? This switch pattern was added not so long ago to avoid code duplication so I'd like to know if it's confusing or missing anything.

OK. I will checkout to another branch and push it for you to check it later!


> I want to keep the Go version in this feature branch. So, I commented out the parts that are required to upgrade Go version when I cherry-pick the code.
> Not in the scope of this PR but I'm for updating Go to the latest version. The current version is already preventing us from upgrading certain packages. cc @donch1989

OK. Thanks for letting me know!
After we check [this topic](https://github.com/instill-ai/component/pull/139#discussion_r1618431995), we can decide if we import the package or keep the current way.

external/langchaingo/textsplitter/token_splitter.go
jvallesm: - About the Go version, I'd import an earlier version of the package that is compatible with our Go version. If this is the reason to copy instead of import the code, a much more sensible approach is updating Go, then importing the latest version of the dependency, even if in the beginning we might potentially lose some functionality like a given token encoding. The version difference is small so this shouldn't be a big migration task.
- About the size of `langchaingo`, the rule of thumb I'd follow would be that it's a good argument as long as what you're trying to copy fits in your package (a function, a struct type). Creating 2 `external` packages and importing them adds confusion to the structure and maybe it's a smell that indicates that it's worth importing the source (and more honest). At this point I don't think binary size is a massive concern and I think there are [plenty of ways to improve it](https://instill-ai.slack.com/archives/C01496MQAQJ/p1714469602131749) anyways. In other SDKs like S3 or Stripe we're using a tiny subset of their features and we're not considering copying their source code.

jvallesm: > Is it different between pipeline-backend-cloud & pipeline-backend now? Should we modify pipeline-backend-cloud?

`pipeline-backend` doesn't use it because this is needed for the Credit feature, implemented only on Cloud.

> I am not quite sure about whether [it is a good source](https://github.com/hupe1980/go-tiktoken).

I picked the other package (without being aware of `tiktoken-go` being used in `component`) because the encoding is embedded by default and because gpt-4o was quickly added after it was announced, so it fit our needs. The star count is definitely an indicator of the quality and health of the project but we shouldn't be guided solely by it and in this case the choice made sense. I'm not against modifying `pipeline-backend-cloud` instead of this package.

operator/text/v0/chunk_text.go
jvallesm: Sorry, I reviewed without having all the context in mind. If this is a workaround to a bug that we can't address in this sprint, I'd add a TODO comment before this line explaining why it's needed and how / when it should be fixed.

operator/text/v0/chunk_text.go
jvallesm: As mentioned, I'm fine keeping this package and updating Pipeline to use it, too. But if we do so, I'd add the offline loader to avoid external calls.

jvallesm:
> split functions belong to external services. If there is an update in external service, the result could be failure.
In my opinion, unit test should mock the external service to focus on what we are testing.
And, the logic of the external service should contain their own test.

@chuang8511 by service you mean we're making external calls to split the text? I.e. not only to packages we import but to online APIs. Is there any call besides the encoding fetch from tiktoken (as mentioned, we can have it offline)? I think this is a task where the results shouldn't change if the dependencies are updated (except on e.g. a bugfix). My fear is that, by mocking the dependencies, we might be hiding the real behaviour of the component and have unexpected results even if the tests pass. Without mocking, a breaking test will signal a breaking change in a dependency. After inspecting the change, we can decide to incorporate the new behaviour if we find it's a legit modification.

I'd start by not mocking the depencencies (it's also less work on our side) and only consider mocking if we find the tests are flickering or requiring frequent updates.

external/langchaingo/textsplitter/token_splitter.go
chuang8511: @jvallesm 
> even if in the beginning we might potentially lose some functionality like a given token encoding. The version difference is small so this shouldn't be a big migration task.

Just share
In my previous experience / industry / company, losing functionality is not acceptable in terms of client sides. So, we barely upgrade the language version to keep the stability.

About Go version, I checked the [release note](https://tip.golang.org/doc/go1.22).
It seems like what you said, it is not a big migration.
However, I still want to separate branches between upgrading Go version & this feature branch.
I think it will be clearer than combining them together.

How about we mark it as TODO and import the package after we upgrade the Go version?
What do you think about it?

>  I'm not against modifying pipeline-backend-cloud instead of this package.

I am confused about this. So, I want to confirm it again.
If we are going to import all langchaingo package, should we modify `pipeline-backend-cloud` to avoid two token package rather than `pipeline-backend`?

chuang8511:
@jvallesm 

> My fear is that, by mocking the dependencies, we might be hiding the real behaviour of the component and have unexpected results even if the tests pass. 

Before read
All contents below are my previous experience, but I think it is highly related to the character of the industry.

We divide unit test & integration test.
By integration test, we can monitor the dependencies.

We separate unit test & integration test because the integration test takes way longer time for CI.
The reason why taking longer time matters is related to two factors
- The confidence of how much the test protects our function.
- How urgent we need to release the functions.

I feel the confidence level below.
- unit test -> 99% confidence.
  - It means no matter how urgent it is, we cannot release without passing unit test.
- integration test -> 80% confidence.
  - It means we can come back to fix it later.

However, I missed considering that this external service won't take such long time.
And, the situation is totally different from my previous experience.
So, I think it will be good not to mock the external service here!

chuang8511:
It is quite weird that I added main_test.go but the coverage did not increase....
Not sure how to check it yet.
I have searched in the report but did not find it.

donch1989:
@chuang8511 
I'll upgrade the Golang version today in another PR.

I'm wondering, if we upgrade the Golang version, do we still need to modify any Golang code from LangChainGo? If not, maybe we don't need to copy their code here, we can just import it.

external/langchaingo/textsplitter/token_splitter.go
donch1989: @chuang8511 
I'll upgrade the Golang version today in another PR.

If my understanding is correct, after we upgrade the Golang version, we don't need to modify any Golang code from LangChainGo, right? If so, let's just import it.

[It has its own data model and interfaces for connecting LLMs, database I/O, etc., which is a bit too heavy for a single Instill Component. We probably want to implement the code ourselves instead of importing it as a whole Go package.](https://linear.app/instill-ai/issue/INS-4499/%5Bcomponent%5D%5Btext-operator%5D-add-task-chunk-text-for-more-advanced#comment-fd95adca)
This only happens when you need to make modifications. If we don't update their code, we can just import it. The size of the package is not a problem, I remember that when using `go build`, the compiler will only compile the code you used.

donch1989:
Hi @chuang8511
I've upgraded the Go version.

external/langchaingo/textsplitter/token_splitter.go
donch1989: I just checked https://github.com/pkoukk/tiktoken-go, and they have added GPT-4 support. I guess we can safely switch to it.

donch1989 (approved):
@chuang8511
I'll merge this PR first. Please note the improvement items, and we can optimize them afterward.

PR #143, opened 2024-05-29
feat: optimise ux for slack component

chuang8511:
Because

- Now, there are several weird parts in Slack component.

This commit

- add user name in read task output, take out is_public_channel for TASK, and change line when posting message.


donch1989:
@chuang8511 
Can you describe the weird parts in the commit message?

application/slack/v0/api_functions.go
```diff
@@ -8,9 +8,9 @@ import (
 	"github.com/slack-go/slack"
 )
 
-func loopChannelListAPI(e *execution, isPublic bool, channelName string) (string, error) {
+func loopChannelListAPI(e *execution, channelName string) (string, error) {
 	var apiParams slack.GetConversationsParameters
-	setChannelType(&apiParams, isPublic)
+	apiParams.Types = append(apiParams.Types, "private_channel", "public_channel")
```
donch1989, line 13: If the `types` is always "private_channel" and "public_channel," we can just make it be constant when sending the request to the Slack API in `GetConversationsForUserContext()`.

application/slack/v0/task_functions.go
```diff
@@ -81,6 +117,40 @@ func (e *execution) readMessage(in *structpb.Struct) (*structpb.Struct, error) {
...
+
+	userIDs = removeDuplicateUserIDs(userIDs)
+	users, err := e.client.GetUsersInfo(userIDs...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	for i, conversation := range readTaskResp.Conversations {
+		for _, user := range *users {
```
donch1989, line 139: We can create a `map[string]string` to convert id to name, which will speed things up.

donch1989 (changes requested):

application/slack/v0/api_functions.go
chuang8511: @donch1989 
Do you mean like these?
1. fork Slack package
2. modify the Slack package
3. Change the import source

chuang8511: Except this, others are done. So, I request the review again!

donch1989: Oh.. sorry, I wrote the wrong function name.
I mean you can have a global variable and use it like this

```
var types []string{"private_channel", "public_channel"}

---

apiParams.Types = types
```

donch1989 (approved):

PR #155, opened 2024-06-06
feat: add read task for gcs

chuang8511:
Because

- we do not have read objects and create bucket task for gcs

This commit

- add read task for gcs
- add create bucket task for gcs

### DIFF
--- a/data/googlecloudstorage/v0/read_objects.go
+++ b/data/googlecloudstorage/v0/read_objects.go
@@ -1,0 +1,49 @@
+package googlecloudstorage
+
+type ReadInput struct {
+	BucketName               string
+	Delimiter                string
+	Prefix                   string
+	Versions                 bool
+	StartOffset              string
+	EndOffset                string
+	IncludeTrailingDelimiter bool
+	MatchGlob                string
+	IncludeFoldersAsPrefixes bool
+}
+
+type ReadOutput struct {
+	TextObjects     []TextObject
+	ImageObjects    []ImageObject
+	DocumentObjects []DocumentObject
+}
+
+type TextObject struct {
+	Data       string
+	Attributes Attributes
+}
+
+type ImageObject struct {
+	Data       string
+	Attributes Attributes
+}
+
+type DocumentObject struct {
+	Data       string
+	Attributes Attributes
+}
+
+type Attributes struct {
+	Name               string
+	ContentType        string
+	ContentLanguage    string
+	Owner              string
+	Size               int64
+	ContentEncoding    string
+	ContentDisposition string
+	MD5                []byte
+	MediaLink          string
+	Metadata           map[string]string
+	StorageClass       string
+}
+
--- a/data/googlecloudstorage/v0/read_objects.go
+++ b/data/googlecloudstorage/v0/read_objects.go
@@ -23,38 +23,50 @@ type ReadInput struct {
 }
 
 type ReadOutput struct {
-	TextObjects     []TextObject
-	ImageObjects    []ImageObject
-	DocumentObjects []DocumentObject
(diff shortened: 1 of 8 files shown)


data/googlecloudstorage/v0/read_objects.go
```diff
@@ -47,3 +57,80 @@ type Attributes struct {
...
+				Attributes: attribute,
+			}
+			output.TextObjects = append(output.TextObjects, textObject)
+		} else if strings.Contains(attrs.ContentType, "image") {
+			imageObject := ImageObject{
+				Data:       string(b),
+				Attributes: attribute,
+			}
+			output.ImageObjects = append(output.ImageObjects, imageObject)
+		} else { // TODO chuang8511: discuss with reviewer what types should we specify here?
```
tillknuesting, line 126: the else is redundant

data/googlecloudstorage/v0/read_objects.go
```diff
@@ -47,3 +57,80 @@ type Attributes struct {
...
+
+	it := client.Bucket(bucketName).Objects(ctx, query)
+
+	output := ReadOutput{
+		TextObjects:     []TextObject{},
+		ImageObjects:    []ImageObject{},
+		DocumentObjects: []DocumentObject{},
+	}
+	for {
+		attrs, err := it.Next()
```
tillknuesting, line 82: Perhaps you could try out to run this concurrently with Goroutines.

chuang8511: After checking these the [google wiki](https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines) & [a discussion thread](https://stackoverflow.com/questions/63215767/google-cloud-storage-listing-objects-concurrently-go#comment111785255_63215767), it seems a better way is to separate the query when there are too many objects to query.

data/googlecloudstorage/v0/read_objects.go
chuang8511: Do you mean it should be like this?
But, if there is no else, all objects will go to DocumentObjects, which is not expected.
Or, do I misunderstand anything?

```
		if strings.Contains(attrs.ContentType, "text") {
			textObject := TextObject{
				Data:       string(b),
				Attributes: attribute,
			}
			output.TextObjects = append(output.TextObjects, textObject)
		} else if strings.Contains(attrs.ContentType, "image") {
			imageObject := ImageObject{
				Data:       string(b),
				Attributes: attribute,
			}
			output.ImageObjects = append(output.ImageObjects, imageObject)
                }

                	documentObject := DocumentObject{
				Data:       string(b),
				Attributes: attribute,
		}
		output.DocumentObjects = append(output.DocumentObjects, documentObject)
```

chuang8511:
@donch1989 
About changing all fields to kebab-case, could I change in this ticket?
Or, do you plan to do it in other ways? (same question for this [PR](https://github.com/instill-ai/component/pull/156).

donch1989:
Hi @chuang8511 
I guess you forgot to upload the definition and `tasks.json`.

chuang8511:
@donch1989 
Sorry, do I miss to read something?
I think it is push. (please correct me if I misunderstand)
🙇 

image

chuang8511:
I will notify you again after I rebase!
I may do it after 3~4 p.m. London time.
Sync to you first.

chuang8511:
same status like [this branch](https://github.com/instill-ai/component/pull/156#issuecomment-2173828568)
Will come back to do final e2e test after the frontend fixes the bug!

chuang8511:
Sync
I have done end to end test.

To make GCS more flexible, I add all params in the output of the component.
However, it damages UI a bit.

image

PR #156, opened 2024-06-10
feat: add read task in bigquery component

chuang8511:
Because

- we want to read from data components

This commit

- add read task to bigquery


chuang8511:
It is only for sync the task progress.
Will change PR status to open later.

data/bigquery/v0/main.go
```diff
@@ -145,3 +149,137 @@ func (c *component) Test(sysVars map[string]any, setup *structpb.Struct) error {
...
+
+		}
+	}
+
+	return def, nil
+}
+
+func constructTableColumns(myDataset *bigquery.Dataset, ctx context.Context, compConfig *base.ComponentConfig) ([]TableColumns, error) {
+	tableIT := myDataset.Tables(ctx)
+	tables := []TableColumns{}
```
donch1989, line 223: How about we only fetch the table we want to use? 

If we still want to fetch multiple tables, we can use a map with table_name as the key.

chuang8511: @donch1989 
I was thinking to take the table name from setting parts to input.
Then, the users can choose the table / tables they want to extract with smart hint or drag-and-drop menu.
And, we also can adjust the columns without saving pipeline again if users want to change table.

What do you think about that?

donch1989: Yeah, sounds good

chuang8511: I have created another ticket and leave TODO memo here.
I will deal with it in the near future.

data/bigquery/v0/main.go
```diff
@@ -122,3 +149,142 @@ func (c *component) Test(sysVars map[string]any, setup *structpb.Struct) error {
...
+}
+
+func constructTableProperties(tables []TableColumns) ([]*structpb.Struct, error) {
+	tableProperties := make([]*structpb.Struct, len(tables))
+
+	for idx, table := range tables {
+		propertiesMap := make(map[string]interface{})
+		for idx, column := range table.Columns {
+			propertiesMap[column.Name] = map[string]interface{}{
+				"title":                column.Name,
```
chuang8511, line 260: TODO: chuang8511
Need to convert other cases to kebab-case.

data/bigquery/v0/insert.go
```diff
@@ -31,8 +32,10 @@ func insertDataToBigQuery(projectID, datasetID, tableName string, valueSaver Dat
 func getDataSaver(input *structpb.Struct, schema bigquery.Schema) (DataSaver, error) {
 	inputObj := input.GetFields()["data"].GetStructValue()
 	dataMap := map[string]bigquery.Value{}
+	transformer := base.InstillDynamicFormatTransformer{}
```
donch1989, line 35: @chuang8511
I don't think we need to convert the case here. Since the schema is not defined in VDP, let's just keep it the same as it is in BigQuery.

donch1989: Yeah, I know. But what if in BigQuery, there are two columns `fooBar` and `foo_bar` at the same time? The logic will go wrong.

donch1989:
Please also help rebase it as well.

data/bigquery/v0/insert.go
chuang8511: So, do you mean that we do not have to transform BigQuery schema into kebab-case for VDP schema?
If so, I will revert this part & [this code](https://github.com/instill-ai/component/pull/156/files#diff-b54962b9f02adac8f1a743deac5ab5f543d79411a4b1c09578d71902aaa6e8a9R266)

donch1989: Yes, I think we can just use the exact same names as those in BigQuery.

chuang8511:
@donch1989 
Sync
Because there is a frontend bug, I have not done the final e2e test.
I will do the test after the Console bug is fixed.
Next time, I will put more time to try hardcode first to skip the bug.
Today, I happen to have a tight schedule, so I could not do it now.

Sorry...

chuang8511:
I have done end-to-end test.
image

PR #157, opened 2024-06-12
chore(proposal): add Anthropic as coming soon AI component

xiaofei-du:
Because

- we want to support Anthropic AI component

This commit

- add skeleton Anthropic AI component (coming soon)

### DIFF
--- a/ai/anthropic/v0/README.mdx
+++ b/ai/anthropic/v0/README.mdx
@@ -1,0 +1,23 @@
+---
+title: "Anthropic"
+lang: "en-US"
+draft: false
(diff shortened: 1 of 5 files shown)


jvallesm (commented):
I assume we want to show this in the website but not in the console. For that we'll need to:
- Set `public: true` so it is hidden from the console.
- Initialize the component so `pipeline-backend` is aware of its existence. For this, the intialized object must comply with the component interface, so an empty package won't be enough. We can create a dummy no-op component that we can import from that package, though. That'll simplify future "coming soon" component additions.

I can address at the end of this sprint, but it'll have to be post freeze code

donch1989:
@xiaofei-du 
Since we already have a pull request for the Anthropic component, I'll close this one.

PR #171, opened 2024-06-21
feat(compogen): add extra section with --extraContents flag'

jvallesm:
Because

- Sometimes we want to extend auto-generated docs with custom content.
  - This is particularly true until JSON schema can be properly parsed by `compogen`, which will add more detail to the complex input / output params.

This commit

- Adds conent-injection capabilities to compogen.

### ℹ️ Info

- You'll notice extra newlines added [everywhere](https://github.com/instill-ai/component/pull/171/files#diff-4b320521a406c97d7c1ccad3dbda779ea82206f7fdba3a7fe69c29d9834f81d0R21). I had avoided this at the cost of a less readable [template](https://github.com/instill-ai/component/pull/171/files#diff-3ddf390734accc49bc9993e2358b740523d552f9701dbadc0e2c1acaec4d97a3R51). But it is ok to have ugly auto-generated code (as long as the rendered Markdown / HTML isn't affected). What we want to read clearly is the source template and the rendered documentation.
- Therefore, review is easier on a commit-by-commit basis 😇 

### Example

https://github.com/instill-ai/component/pull/172 exemplifies how to add extra sections in the Text operator.


ai/archetypeai/v0/README.mdx
```diff
@@ -12,59 +12,96 @@ It can carry out the following tasks:
 - [Summarize](#summarize)
 - [Upload File](#upload-file)
 
+
+
 ## Release Stage
 
 `Alpha`
 
+
```
jvallesm, line 21: ℹ️ You'll notice extra newlines added everywhere. I had avoided this at the cost of a less readable template. But it is ok to have ugly auto-generated code (as long as the rendered Markdown / HTML isn't affected). What we want to read clearly is the source template and the rendered documentation.

PR #172, opened 2024-06-21
chore(compogen): exemplify content injection with Text operator

jvallesm:
- **chore(compogen): update dependencies**
- **feat(compogen): add version to command**
- **feat(compogen): inject extra content in document**
- **feat(compogen): add path to extra content as argument**
- **feat(compogen): add extra content specifying the section**
- **feat(compogen): bump version**
- **feat: document compogen changes in contribution guideline**
- **feat: regenerate READMEs with new compogen**
- **example: extra sections in Text operator**

Because

- (write the reason why we need to consider this PR in a list)

This commit

- (write the summary of all commits in this PR in a list)


operator/text/v0/.compogen/extra-chunk.mdx
```diff
@@ -0,0 +1,10 @@
+The `strategy field defines the chunking method through the `chunk-method`
```
pinglin, line 1:
```suggestion
The strategy field defines the chunking method through the `chunk-method`
```
But it looks there is no `chunk-method` in the document?

jvallesm: From the parent PR

> This is particularly true until JSON schema can be properly parsed by compogen, which will add more detail to the complex input / output params.

In order to have the input / output auto-generated with all the fields, we need to parse and render any JSON schema, not only the fields we know we'll find. This would be the real fix for the problem we have in the text operator, but it would require more effort. As the conversation popped in the middle of the sprint, adding an extra section explaining how this field works is a workaround we could implement without affecting other items in the sprint.

Not that this isn't a proposal for the actual text operator page, I just added some placeholder documentation to showcase the required input and the result. Someone closer to the text operator / RAG application should write this new section, or I can tackle it but I'll have to do it after I finish my sprint tasks.

jvallesm:
With https://github.com/instill-ai/component/pull/178 we'll already have a production example.

PR #175, opened 2024-06-25
docs(CONTRIBUTION): align content with the latest documentation

pinglin:
Because

- the `CONTRIBUTION.md` file has typos and inconsistent content with the latest documentation website

This commit

- update the file

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -14,12 +14,12 @@ section](https://github.com/instill-ai/community#contributing) for more details.
 ## Concepts
 
 Before delving into the details to come up with your first PR, please
-familiarize yourself with the project structure of [Instill
-Core](https://github.com/instill-ai/community#instill-core).
+familiarize yourself with the project structure of 🔮 [**Instill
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -56,7 +63,7 @@ various providers, whether it's the primary Instill Model or those from
 third-party AI vendors. They are defined and initialized in the [`ai`](../ai)
 package.
 
-#### Data
+#### `Data`
```
jvallesm, line 66: ❓ Do we want this headers with `code` style? Generic, AI, Data, etc. are end-user names, not IDs, and they just appeared as **bold** in the parent section.

pinglin: I was thinking to stand out those component types from the parent header. If they look weird and breaking the header hierarchy, I can reset it.

jvallesm: To me it was a bit confusing, headers already convey emphasis so it lead me to think the code style meant something else.

jvallesm (approved):

PR #178, opened 2024-06-25
feat: add mail component

chuang8511:
Because

- we want users to be able to fetch mail and send mail through vdp

This commit

- build mail component with send and read mail


chuang8511:
e2e test

read email test
image

send email test
image

application/email/v0/read_emails.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return output, nil
+}
+
+func initIMAPClient(serverAddress string, serverPort int) (*imapclient.Client, error) {
+
+	c, err := imapclient.DialTLS(serverAddress+":"+strconv.Itoa(serverPort), nil)
```
donch1989, line 98: We can use `fmt.Printf` here

application/email/v0/read_emails.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+	}
+	if subject, err := h.Text("Subject"); err != nil {
+		log.Fatalf("failed to get subject: %v", err)
+	} else {
+		email.Subject = subject
+	}
+}
+
+func includeSearchCondition(email Email, search Search) bool {
```
donch1989, line 239: How about rename this to `checkEnvelopeCondition`?

application/email/v0/read_emails.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	if search.Date != "" {
+		if !strings.Contains(email.Date, search.Date) {
+			return false
+		}
+	}
+	return true
+}
+
+func includeSearchMessage(email Email, search Search) bool {
```
donch1989, line 263: How about rename this to `checkMessageCondition`?

application/email/v0/config/tasks.json
```diff
@@ -0,0 +1,471 @@
...
+      "instillUpstreamTypes": [
+        "value",
+        "reference",
+        "template"
+      ],
+      "instillUIOrder": 4,
+      "title": "Search Subject",
+      "type": "string"
+    },
+    "search-from-email": {
```
donch1989, line 17: `search-from-email` looks a little bit weird to me. How about we use `search-from` or `search-email-from` instead?

application/email/v0/config/tasks.json
```diff
@@ -0,0 +1,471 @@
+{
+  "$defs": {
+    "search-subject-text": {
```
donch1989, line 3: I thinks `search-subject` should be enough.

application/email/v0/read_emails.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func includeSearchCondition(email Email, search Search) bool {
+	if search.SearchSubjectText != "" {
+		if !strings.Contains(email.Subject, search.SearchSubjectText) {
+			return false
+		}
+	}
+	if search.SearchFromEmail != "" {
+		if !strings.Contains(email.From, search.SearchFromEmail) {
```
donch1989, line 246: I suggest we perform an exact search here: `email.From == search.SearchFromEmail`. If the email address is `aaa@bbb.cc` but you search for `aa@bbb.cc`, things will go wrong.

application/email/v0/read_emails.go
```diff
@@ -0,0 +1,274 @@
...
+			return nil, err
+		}
+
+		emails = append(emails, email)
+	}
+
+	return emails, nil
+}
+
+func convertMailboxString(mailbox string) string {
```
donch1989, line 200: Does this mean that the current implementation can only be used with Gmail?

application/email/v0/config/tasks.json
```diff
@@ -0,0 +1,471 @@
...
+      "instillUpstreamTypes": [
+        "value",
+        "reference",
+        "template"
+      ],
+      "instillUIOrder": 3,
+      "title": "Search To",
+      "type": "string"
+    },
+    "server-address": {
```
donch1989, line 45: I'm thinking we should put the `server-address` and `server-port` into `setup.json`. They should be in the same place as the user email and password.

donch1989 (changes requested):

application/email/v0/read_emails.go
chuang8511: @donch1989 
I miss to consider this point when I developed it. Thanks for pointing it out.
Now, I have 2 ideas for this.

1. do not support mailbox field, and we fetch all emails from all mailbox.
2. make mailbox as plain text, and we write the Gmail input as the example in Email Component Document

What do you think about this?

donch1989: I think we can go with option 1 first. Will that have any side effects?

chuang8511: > Will that have any side effects?

It will take a longer time to scan all mailboxes.
For AI SDR demo, I think the speed of fetching emails & accuracy of the content would be the key to success.
By specifying the mailbox, it can increase the performance in terms of speed and accuracy.

application/email/v0/.compogen/extra-intro.mdx
```diff
@@ -0,0 +1,3 @@
+You can connect different email server through Email component.
+We use SMTP protocol to send emails, So, you can set the SMTP server address and port in the component configuration.
+We use IMAP protocol to fetch emails, So, you can set the IMAP server address and port in the component configuration.
```
jvallesm, line 3: `s/So/so`

As a small suggestion, let's try to avoid repetition:
```suggestion
Emails are fetched and sent using the IMAP and SMTP protocols, respectively. You can set the server address and port for each protocol in the component configuration.
```

application/email/v0/.compogen/extra-send-mails.mdx
```diff
@@ -0,0 +1,9 @@
+Supported Provider Samples
+
+| Provider | SMTP Server Address | Port | Description |
+| --- | --- | --- |
```
jvallesm, line 4:
```suggestion
| --- | --- | --- | --- |
```

application/email/v0/.compogen/extra-send-mails.mdx
```diff
@@ -0,0 +1,9 @@
+Supported Provider Samples
+
+| Provider | SMTP Server Address | Port | Description |
```
jvallesm, line 3:
```suggestion
| Provider | SMTP Server Address | Port | Notes |
```

application/email/v0/config/definition.json
```diff
@@ -0,0 +1,19 @@
+{
+  "availableTasks": [
+    "TASK_SEND_EMAILS",
+    "TASK_READ_EMAILS"
+  ],
+  "documentationUrl": "https://www.instill.tech/docs/component/application/email",
+  "icon": "assets/email.svg",
+  "id": "email",
+  "title": "Email",
+  "description": "Get and send emails from Mail Protocol",
```
jvallesm, line 10:
```suggestion
  "id": "email",
  "public": true,
  "title": "Email",
  "description": "Get and send emails from Mail Protocol",
```

application/email/v0/read_emails.go
chuang8511: My original design is to let the users able to search `aaa@bbb.cc` through `aaa`.
But, I think your way will be better! So, I will modify it as well.

application/email/v0/read_emails.go
chuang8511: Could we go with option 2 in the current phase?
Is only searching one mailbox a damage? If so, how about we set it as array:string?
Or, do you have other concerns that I miss?

donch1989: > It will take a longer time to scan all mailboxes.

I'm not sure about the mechanism here. I remember you already set a limit on the number of emails, right?

> array:string
I don't think array: string can work well on Console for now.

The problem with Option 2 is that users may struggle to type the correct mailbox name.

application/email/v0/read_emails.go
donch1989: Yeah, in this case, fuzzy search it not suitable.

application/email/v0/read_emails.go
chuang8511: > I'm not sure about the mechanism here. I remember you already set a limit on the number of emails, right?

Yes, the limit I set is the number to "scan", it will only scan a specific mailbox if the limit is smaller than the message number in a mailbox.
By the current design, it won't get the content for AI SDR usage if we make it scan INBOX first.

The purpose I specially design mailbox field is for the usage to get the sending mails specifically.

> The problem with Option 2 is that users may struggle to type the correct mailbox name.

Yes, it is the purpose to design this field with enum.

chuang8511: > the limit I set is the number to "scan",

The reason of this design is that I have not found how to fetch emails through batch way.
So, we have to fetch email [one by one ](https://github.com/instill-ai/component/pull/178/files#diff-0382a2917f426e2ffd3a03998b35a9198571869579d4806de3a03eec10b16caeR129).

It makes the scanning takes time if we do not have limit here.
So, I added this field and set it 100 as default.

donch1989: OK, I just check the imap package, it need the mailbox input. Then I think we can use Option 2 with a string plaintext.

chuang8511: Thank you! I will add Gmail mailbox in the document as well.

application/email/v0/config/tasks.json
```diff
@@ -0,0 +1,329 @@
...
+        "value",
+        "reference",
+        "template"
+      ],
+      "instillUIOrder": 7,
+      "title": "Search Message",
+      "type": "string"
+    }
+  },
+  "TASK_SEND_EMAILS": {
```
donch1989, line 88: Since we only send one email, how about removing the "s"?

donch1989 (changes requested):

application/email/v0/.compogen/extra-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+| Outlook | imap-mail.outlook.com | 993 |
+| office365 | outlook.office365.com | 993 |
+| Yahoo | imap.mail.yahoo.com | 993 |
+| iCloud | imap.mail.me.com | 993 |
+
+
+For App Password, please follow the steps below:
+ - 1. Please Sign in to your Google Account with link: https://myaccount.google.com/apppasswords
+ - 2. Create a new App Password and save it in a secure place.
+ - 3. Set it in Secrets in the Instill Platform.
```
GeorgeWilliamStrong, line 26:
```suggestion
 - 3. Add App Password as a new secret in the Instill Platform by navigating to **Console** > **Settings** > **Secrets**.
```

GeorgeWilliamStrong (approved):
LGTM (from a doc perspective)

application/email/v0/config/tasks.json
chuang8511: Ah, I thought an email is like a message to a recipient.
The emails mean a message to several recipients.

I set it as [recipients](https://github.com/instill-ai/component/pull/178/files#diff-f5135e1e71de2a32c9a05b4a2876cb66e90216e330f0fdbcebc2d6f81b6b1079R94)

So, I use plural here.
Should I change it to singular?

donch1989: 'Send emails' sounds more like sending different content, but in this case, you're sending just one.

donch1989 (approved):
LGTM

PR #197, opened 2024-07-04
feat: unify pipeline and component usage handlers

jvallesm:
Because

- We have usage handlers for pipeline and component but the former only receives the number of components. We can merge them and simplify our logic.
- When a component is configured to use (secret) credentials injected from the environment variables in the configuration (e.g. when a global OpenAI API key is set), the component logic always references these values as global secrets, and sometimes just "secrets". This is confusing because users can define their own secrets. "Instill Credentials" is more concrete and less confusing.
- The `GetID()` method in the `IComponent` interface actually refers to the component definition ID. At some point we'll actually want to know the component ID (i.e. the component in the pipeline `mycomponent-1` vs the definition `mycomponent`), e.g. for pipeline trigger logging.

This commit

- Simplifies the usage handler logic. Instead of receiving a map (component def ID->usage handler), a single handler is received, which is responsible of accessing the definition ID and apply the relevant logic.
- Names for global secrets and component definition IDs in the component and execution interfaces are updated. We don't implement the ID accessor (yet) but we set the proper name for the definition ID getter.

### DIFF
--- a/ai/anthropic/v0/component_test.go
+++ b/ai/anthropic/v0/component_test.go
@@ -9,7 +9,6 @@ import (
 	"testing"
 
 	qt "github.com/frankban/quicktest"
-	"go.uber.org/zap"
 	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/instill-ai/component/base"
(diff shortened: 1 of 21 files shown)


donch1989 (approved):
LGTM.
I'll merge this directly, and we can rebase other PRs to catch up with the changes.

PR #198, opened 2024-07-05
feat: add mongodb component

zidanehakim:
Because

- We need to integrate MongoDB into our data pipeline.
- This will allow us to cover data movement user stories

This commit

- Allow client to connect to their MongoDB database through VDP
- Mongo URI auth
- Create execution function to handle these following tasks

- [x] TASK_INSERT
- [x] TASK_UPDATE
- [x] TASK_FIND
- [x] TASK_DELETE
- [x] TASK_DROP_COLLECTION
- [x] TASK_DROP_DATABASE
- [x] TASK_INSERT_MANY


data/mongodb/v0/config/tasks.json
```diff
@@ -0,0 +1,457 @@
...
+          "instillUIOrder": 2,
+          "instillUpstreamTypes": [
+            "reference",
+            "template",
+            "value"
+          ],
+          "title": "Criteria"
+        },
+        "limit":{
+          "description": "The number of documents to return, if 0 then all documents will be returned, otherwise the number of documents to return",
```
chuang8511, line 127: Question

> if 0 then all documents will be returned, otherwise the number of documents to return",

How about we just make it optional & judge it in our codebase rather than declaring it in document?

0 turns out to be all seems not intuitive.
Null turns out to be all make more sense in my opinion.

data/mongodb/v0/config/tasks.json
```diff
@@ -0,0 +1,457 @@
...
+        "documents": {
+          "description": "The documents returned from the find operation",
+          "instillEditOnNodeFields": [],
+          "instillUIOrder": 0,
+          "required": [],
+          "title": "Documents",
+          "type": "array",
+          "instillFormat": "array:semi-structured/json",
+          "items": {
+            "title": "Result",
```
chuang8511, line 171: nit
documents contain document.
Just like when we write a loop, we do it like

```
for _, document := range documents {
...
...
}
```

data/mongodb/v0/config/tasks.json
```diff
@@ -0,0 +1,457 @@
...
+          "instillUIOrder": 1,
+          "instillUpstreamTypes": [
+            "reference",
+            "template",
+            "value"
+          ],
+          "title": "Collection Name",
+          "type": "string"
+        },
+        "criteria": {
```
chuang8511, line 218: Same question as Find one.

data/mongodb/v0/config/tasks.json
```diff
@@ -0,0 +1,457 @@
...
+          "instillShortDescription": "JSON Data",
+          "instillUIOrder": 2,
+          "instillUpstreamTypes": [
+            "reference",
+            "template",
+            "value"
+          ],
+          "title": "Criteria"
+        },
+        "update": {
```
chuang8511, line 232: nit
`update-data` better?

data/mongodb/v0/config/tasks.json
```diff
@@ -0,0 +1,457 @@
...
+          "instillUIOrder": 1,
+          "instillUpstreamTypes": [
+            "reference",
+            "template",
+            "value"
+          ],
+          "title": "Collection Name",
+          "type": "string"
+        },
+        "criteria": {
```
chuang8511, line 309: Same question above

data/mongodb/v0/config/definition.json
```diff
@@ -0,0 +1,26 @@
+{
+  "availableTasks": [
+    "TASK_INSERT",
+    "TASK_FIND",
+    "TASK_UPDATE",
+    "TASK_DELETE",
+    "TASK_DROP_COLLECTION",
+    "TASK_DROP_DATABASE"
```
chuang8511, line 8: question
What use case or design for to add  `"TASK_DROP_COLLECTION",    "TASK_DROP_DATABASE"`

chuang8511: And, if we can drop collection & database, how about we also can create collection & database?

data/mongodb/v0/main.go
```diff
@@ -0,0 +1,144 @@
...
+	FindOne(ctx context.Context, filter interface{}, opts ...*options.FindOneOptions) *mongo.SingleResult
+	InsertOne(ctx context.Context, document interface{}, opts ...*options.InsertOneOptions) (*mongo.InsertOneResult, error)
+	UpdateMany(ctx context.Context, filter interface{}, update interface{}, opts ...*options.UpdateOptions) (*mongo.UpdateResult, error)
+	DeleteMany(ctx context.Context, filter interface{}, opts ...*options.DeleteOptions) (*mongo.DeleteResult, error)
+	Drop(ctx context.Context) error
+}
+
+type MongoDBClient interface {
+	Drop(ctx context.Context) error
+}
```
chuang8511, line 53: question
Is there a concern that you divide MongoCollectionClient & MongoDBClient?
They seem to be under MongoDBClient.
And, this DB does not mean database but it is the name of MongoDB.

e.g. If the MongoDB name with Mongo, then it could be like
```
type Mongo interface {
  Find()
  FindOne()
...
  DropDatabase()
  CreateDatabase()
...
}
```

data/mongodb/v0/tasks.go
```diff
@@ -0,0 +1,280 @@
...
+		projection := bson.M{}
+		projection["_id"] = 0
+		for key := range criteria {
+			projection[key] = 1
+		}
+		findOptions.SetProjection(projection)
+
+		// Find all documents with the specified options (including limit if set)
+		cursor, err = e.collectionClient.Find(ctx, realCriteria, findOptions)
+	}
```
chuang8511, line 130: Just my preference. You can just take a look to check if they are same and neater.


```
if len(criteria) != 0 {
		projection := bson.M{}
		projection["_id"] = 0
		for key := range criteria {
			projection[key] = 1
		}
		findOptions.SetProjection(projection)
}
cursor, err = e.collectionClient.Find(ctx, realCriteria, findOptions)

```
	if len(criteria) == 0 {
		// Find all documents with the specified options (including limit if set)
		cursor, err = e.collectionClient.Find(ctx, realCriteria, findOptions)
	} else {
		// Define the projection to include only the specified fields
		projection := bson.M{}
		projection["_id"] = 0
		for key := range criteria {
			projection[key] = 1
		}
		findOptions.SetProjection(projection)

		// Find all documents with the specified options (including limit if set)
		cursor, err = e.collectionClient.Find(ctx, realCriteria, findOptions)
	}
```

File: data/mongodb/v0/tasks.go [hunk 9]
```diff
@@ -0,0 +1,280 @@
...
+
+	var documents []map[string]any
+	for cursor.Next(ctx) {
+		var row map[string]any
+		err := cursor.Decode(&row)
+		if err != nil {
+			return nil, err
+		}
+		documents = append(documents, row)
```
chuang8511, line 143: nit
It seems weird that documents append row but not document.
`documents = append(documents, document)`
`documents = append(documents, row)`

data/mongodb/v0/tasks.go
chuang8511: And another opinion & preference.
Code should explain itself. So, I usually do not add comments to explain what code does but why code is written like it.

e.g. [Sample I recently did](https://github.com/instill-ai/component/blob/main/operator/text/v0/markdown_splitter.go#L90)
It is not a perfect sample. I just want to share my habit.
If you think it is a good point, you can follow it. It will be your call!

data/mongodb/v0/tasks.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	if len(unsetFields) > 0 {
+		updateDoc["$unset"] = unsetFields
+	}
+
+	if len(updateDoc) == 0 {
+		return nil, fmt.Errorf("no valid update operations found")
+	}
+
+	_, err = e.collectionClient.UpdateMany(ctx, criteria, updateDoc)
```
chuang8511, line 202: question
I saw in json schema, you set update for 1 data.
But why here is `UpdateMany`?

chuang8511 (commented):
Hi @zidanehakim 
Thanks for your work!
I leave some comments for product & coding.
Some of them are opinions & preference, please take a look.

Let's discuss them!

data/mongodb/v0/config/tasks.json
```diff
@@ -0,0 +1,457 @@
...
```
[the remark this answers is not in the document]
zidanehakim: > Question Does it make sense to call it filter? [source](https://www.mongodb.com/docs/manual/tutorial/query-documents/) image

Yes, I feel like filter sounds better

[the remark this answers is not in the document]
zidanehakim: @chuang8511

Yes, I also feel that "filter" fits better than "criteria." Will change it soon!

Regarding the use of SQL commands with NoSQL databases, in my opinion, it’s better to use the corresponding database standard and language. Some people may only be familiar with MongoDB, which could be why they choose it and then SQL-like structure input comes out of nowhere doesn't seem to be a good UX for me. In this case, MongoDB uses "aggregate" for filtering and takes JSON or struct as its input.

Additionally, using SQL queries directly may harm the database such as "SQL Injection." For example, if we have a query like "SELECT * FROM inventory WHERE {user input}," the user input could be manipulated into something like "username = '' OR 1=1; DROP TABLE inventory;" which can delete all the data in the table inventory. This is the trade off I have been considering, even when deciding to use JSON for SQL component rather than SQL query. Please your advice on this matter.

[sql injection](https://medium.com/@craftingcode/understanding-sql-injection-attacks-and-how-to-prevent-them-8750ff1384fd)

data/mongodb/v0/config/definition.json
zidanehakim: @chuang8511

I'm thinking of when the user wants to migrate from one database to another, it could be the whole database or just some specific collection, and when the process has finished, they may want to delete the old data source so that might be the use case.

As for create collection & database, `insert` method from mongodb sdk provide an auto create feature if either collection or database doesn't exist. Moreover, mongodb have flexible structure unlike sql, in a single collection, it may have various fields for each documents. So I guess there's no need to add create collection and database.

data/mongodb/v0/main.go
zidanehakim: @chuang8511 

This is the part where I would like your suggestions the most, the reason why I divide it into two structs is so that it could be use for mock input for component test as well.

The structure of mongodb go sdk looks something like

`type mongo.Client interface {
  Database(...) mongo.Database
}

type mongo.Database interface {
  Collection(...) mongo.Collection
  Drop(...) error
}

type mongo.Collection interface {
 Find()
  FindOne()
...
  DropDatabase()
  CreateDatabase()
  Drop()
}`

It sort of embed the struct hierarchy 2 levels before we can access the main function on mongo.Collection, moreover there's also TASK_DROP_DATABASE where we don't need it continue up until mongo.Collection. So I had decided to seperate it into two client (MongoDatabaseClient, MongoCollectionClient) where MongoDatabaseClient will be used only for TASK_DROP_DATABASE. I'm not sure whether it's a good implementation, please your suggestions.

data/mongodb/v0/tasks.go
zidanehakim: @chuang8511 

> I saw in json schema, you set update for 1 data.

Sorry but I would like to know which json schema is it ? I always intend to use UpdateMany since it can be used for both single or multiple document. Field like `"city":"New York"` on criteria/filter may have multiple documents, and updated to specific structure. We could also make use of mongodb `_id` if we want one specific document combined with another field.

data/mongodb/v0/config/tasks.json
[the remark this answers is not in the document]
chuang8511: > In this case, MongoDB uses "aggregate" for filtering and takes JSON or struct as its input.

OK, I do not have much experience with MongoDB. If aggregate makes more sense in community, we should use it!

data/mongodb/v0/config/definition.json
chuang8511: > they may want to delete the old data source so that might be the use case.

Seems good in terms of cost.

> insert method from mongodb sdk provide an auto create feature if either collection or database doesn't exist.

OK! Good!

data/mongodb/v0/main.go
chuang8511: In this MongoDB case, I think it would be fine because it does not influence too much with only 2 layers.

Some engineers would like to create interface as much as possible for the future space.
But, in my opinion, I do not like over design because it increases the cost of maintenance.


Abstractly, I think coding is like creating a world and designing code structure is like grouping concept.

Let's take animal as example.
At first, the world is small, you only have dog & cat. 
So, you will only need an interface called animal.
But, when your world become bigger, there may be more kinds of cats, you may need another interface called cat for other types of cats.
However, if your world is unbalanced, there is only a dog but thousands of cat. The interface animal may not so necessary. Then, we can consider to take it off to simplify the world.
The story will be too long if I kept writing. So, I stop here.

If you are interested in this kind of concept, you can read design patterns. I feel it is like philosophy.
And, my belief is "as long as we think the code deeply with philosophy, we can increase the maintenance of codebase".

data/mongodb/v0/tasks.go
chuang8511: @zidanehakim 
```
        "update-data": {
          "description": "The updated data to be applied to the documents",
          "instillAcceptFormats": [
            "semi-structured/*","structured/*","object"
          ],
          "instillUIOrder": 3,
          "instillUpstreamTypes": [
            "reference",
            "template",
            "value"
          ],
          "title": "Update"
        }
```

I thought it should be array:object, array:semi-structured/*, ... right?
I am not so sure if object contains array concept.

data/mongodb/v0/main.go
chuang8511: However, I read the code here.
Probably your design makes you need to do it in main file.
If you only have a layer, you can make this logic in the taskFunction. right?
Then, you won't need to do `if e.Task != TaskDropDatabase && e.client.collectionClient == nil {` this kind of logic.

```
		var db *mongo.Database
		if e.client.databaseClient == nil {
			db = client.Database(inputStruct.DatabaseName)
			e.client.databaseClient = db
		}

		if e.Task != TaskDropDatabase && e.client.collectionClient == nil {
			collection := db.Collection(inputStruct.CollectionName)
			e.client.collectionClient = collection
			e.client.searchIndexClient = collection.SearchIndexes()
		}
```

chuang8511: However, I read the code here.
Probably your design makes you need to do it in main file.
If you only have a layer, you can make this logic in the taskFunction. right?
Then, you won't need to do if e.Task != TaskDropDatabase && e.client.collectionClient == nil { this kind of logic.

```
		var db *mongo.Database
		if e.client.databaseClient == nil {
			db = client.Database(inputStruct.DatabaseName)
			e.client.databaseClient = db
		}

		if e.Task != TaskDropDatabase && e.client.collectionClient == nil {
			collection := db.Collection(inputStruct.CollectionName)
			e.client.collectionClient = collection
			e.client.searchIndexClient = collection.SearchIndexes()
		}
```

data/mongodb/v0/tasks.go
```diff
@@ -0,0 +1,437 @@
...
+				},
+			},
+		},
+	}
+
+	if len(fields) > 0 {
+		query = append(query, bson.M{
+			"$project": project,
+		})
+	}
```
chuang8511, line 409: Important
About the logic above, I am not so sure the logic about this part.
```
	if !exact {
		if numCandidates > 0 {
			vectorSearch["numCandidates"] = numCandidates
		} else {
			vectorSearch["numCandidates"] = 3 * limit
		}
	}

	project := bson.M{"_id": 0}
	for _, field := range fields {
		project[field] = 1
	}

```

1. when exact is true, what should it be for numCandidates? 
2. why do we need to set `project[field] = 1`? What does it mean here?

I am not so familiar with this part, please make sure the function works as expectation!

chuang8511 (commented):
Hi @zidanehakim 
I left a few comments including the previous review.
Please take a look.
In general, I think others LGTM.

data/mongodb/v0/tasks.go
zidanehakim: @chuang8511 

1. According to mongodb atlas vector search docs,`exact :false - to run ANN search, true - to run ENN search`. Basically ANN doesn't consider every data and uses some efficient algorithm instead, which is pretty good for larger scale collection (> 10000 documents is preferable according to mongodb, that's why numCandidates is needed. Whilst ENN search consider every data within a collection (<= 10000 documents is preferable), numCandidates isnt needed as every documents is being considered within the vector search.

2. In mongodb aggregate, if we want to filter the retrieved documents with specific fields, we have to construct map that looks like
`"$project":{
"field1" : 1,
"field2" : 1
}`
where 1 means true and 0 means false

data/mongodb/v0/tasks.go
zidanehakim: @chuang8511

Oh, my intention is for instance if we have 3 documents with same city and different name,
`{"name":"a","b","c", "city":"New York"}`

For filter we can put,
`{"city":"New York"}`

For update-data we can put,
`{"city":"Washington"}`

Then all those 3 documents change into
`{"name":"a","b","c", "city":"Washington"}`


So both filter and update-data has to be of type object, for multiple update-data the user might use iterator.

data/mongodb/v0/main.go
zidanehakim: @chuang8511

> But, when your world become bigger, there may be more kinds of cats, you may need another interface called cat for other types of cats.

Good point, in the end I've decided to keep MongoCollectionClient and MongoDatabaseClient type for component_test.go, but to group it up as to unify those two types (since both are part of mongo.Client), i've decided to make one more type

```
type MongoClient struct {
	collectionClient  MongoCollectionClient
	databaseClient    MongoDatabaseClient
	searchIndexClient MongoSearchIndexClient (This one is new, for vector search)
}
```

so I dont need to put if e.Task != TaskDropDatabase anymore and let the unused client be of nil value.

However, for both e.client.databaseClient == nil and e.client.collectionclient are needed for component_test.go since it uses mock client and preventing it to be overwritten

data/mongodb/v0/tasks.go
zidanehakim: @chuang8511 

And one more thing is that I put _id : 0 in order not to show id field by default, since we talk about data migration, sometimes the user doesn't really need such metadata, but still if the user insist on including _id, they still can do it via fields input, for instance ["name", "age", "_id"]

PR #200, opened 2024-07-05
feat(openai): add dimensions in openai component

chuang8511:
Because

- in many use cases, users may not need to output vectors of such high dimensionality and might only require the first 'x' dimensions.

This commit

- add dimensions in api caller

### DIFF
--- a/ai/openai/v0/config/tasks.json
+++ b/ai/openai/v0/config/tasks.json
@@ -173,6 +173,19 @@
           ],
           "title": "Text",
           "type": "string"
+        },
+        "dimensions": {
+          "description": "The number of dimensions the resulting output embeddings should have. Only supported in text-embedding-3 and later models.",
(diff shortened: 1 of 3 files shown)


chuang8511:
Will need to do end to end test before converting to Open

chuang8511:
Tested
image

donch1989:
@chuang8511
Can you confirm it still works if we don't set the dimension and leave it empty?

chuang8511:
@donch1989 
Thanks for pointing this. I missed it.
I did not find the official document for the default dimensions. 
So, I did not set it by default but keep the original ways.

E2E test
image
image

donch1989 (approved):
LGTM.
By the way, please use `feat(component-id): xxx` for the PR title next time.
For example:
From `feat: add dimensions in openai component` to `feat(openai): add dimensions in openai component`.

PR #201, opened 2024-07-08
feat: add JQ input field that accepts any type

jvallesm:
Because

- The `jq` task took only strings, which can be leveraged for reference but forced users to pre-process objects with another JSON operator that marshalled the input.

This commit

- Introduces a new input field for the task, which can be any object
  - The old value is hidden under the "More" settings section. If the new field is empty, task will fall back to the old field (for backwards compatibility).
- Removes object type assumption in unmarshal task: now it can unmarshal any valid JSON value.

### DIFF
--- a/operator/json/v0/README.mdx
+++ b/operator/json/v0/README.mdx
@@ -80,8 +80,9 @@ Process JSON through a `jq` command
 | Input | ID | Type | Description |
 | :--- | :--- | :--- | :--- |
 | Task ID (required) | `task` | string | `TASK_JQ` |
-| JSON input (required) | `jsonInput` | string | JSON string to be processed |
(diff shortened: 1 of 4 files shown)


jvallesm:
# QA

## Before

CleanShot 2024-07-11 at 17 22 40

## After

### No pre-processing required

CleanShot 2024-07-11 at 17 25 00

### Reference can be achieved by marshalling

CleanShot 2024-07-11 at 17 27 21

GeorgeWilliamStrong:
Looks great - thanks @jvallesm. Sorry, I didn't have time to review this until just now 🫠

GeorgeWilliamStrong (commented):
LGTM

PR #202, opened 2024-07-08
feat(instill): send requester UID, if present, on model trigger

jvallesm:
Because

- Instill Model will [use the requester UID,](https://github.com/instill-ai/model-backend/pull/619/files) when present, in model execution.

This commit

- Adds the requester header to the trigger model call in the `instill` component.


donch1989 (approved):
LGTM

PR #209, opened 2024-07-10
feat(text): add input and output and fix bugs

chuang8511:
Because

- position in the chunks are different from the original source
- model name and encoding name are not inter-changeable
- markdown chunk and recursive chunk need the token count as well
- document for text component is not clear
- precommit hook modify the test data which cause the test case failed

This commit

- fix position bugs
- add model name and token count to markdown chunk and recursive chunk
- take out encoding name in token chunk
- use injection to make the text document clear
- fix precommit hook


donch1989 (approved):
LGTM

PR #210, opened 2024-07-11
feat: add handler to auto-fill missing default values

YCK1130:
Because:
 - Request from frontend and direct API calls may miss some optional fields
 - The underlying meaning of default is that if no value is provided, we will use this value instead.

This commit

- Implement a function in the `base` package for all components to use
- The function supports filling fields with primitive types, including string, integer, number, boolean, and array of them.
- Support allOf, anyOf, oneOf fields
- Recursively fill in multi-layer object value

### DIFF
--- a/application/github/v0/main.go
+++ b/application/github/v0/main.go
@@ -4,7 +4,6 @@ package github
 import (
 	"context"
 	_ "embed"
-	"encoding/json"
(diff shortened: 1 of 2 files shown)


donch1989 (approved):
LGTM

PR #212, opened 2024-07-11
feat: GitHub component pagination

YCK1130:
Because

- Users may want to control how much data they wish to receive.

This commit

- Add pagination options to all list tasks
- Update readme


donch1989 (approved):
LGTM

PR #213, opened 2024-07-12
feat(cohere): add cohere to be able to use instill credit

chuang8511:
Because

- we should expose users' input & output for instill credit calculation
- The logic between cloud & local should align

This commit

- expose users' input & output
- refactor the embedding function to make the business logic function exposed


chuang8511:
Test result is in the ticket.

PR #223, opened 2024-07-17
chore: improve JSON operator documentation

jvallesm:
Because

- JSON operator documentation could be clearer.

This commit

- Replaces "JSON object" by "JSON entity" in order to reflect that any JSON datatype is accepted. Technically "object" is valid but it might be confused with the JSON hash type.
- Adds an extra section to document `jq` and share some examples.


donch1989 (approved):
LGTM

PR #229, opened 2024-07-18
feat: add new models in open ai

chuang8511:
Because

- there are new models in openai

This commit

- add new models in openai component


tillknuesting:
lgtm

PR #232, opened 2024-07-22
feat(document): add docx doc pptx ppt html to transform to text in markdown format

chuang8511:
Because

- we want to handle more types of unstructured data

This commit

- add the transformer for docx doc pptx ppt html to markdown

In high level, we do
- judge what file extensions
  - If it is pdf, directly use pdfplumber to convert pdf to markdown
  - If it is docx / doc / ppt / pptx, transform the document to pdf by libreoffice, convert pdf to markdown with pdfplumber.
  - If it is html, transform it to markdown with html-to-markdown package

The reason we separate docx/doc & ppt/pptx is they are essentially different for users according to use cases.
It will differentiate the pdf to markdown logic. So, we divide them first.


chuang8511:
Will need to deal with [this issue](https://github.com/instill-ai/component/pull/238#discussion_r1692736843) in this PR as well.
 Please wait for me.

chuang8511:
fixed

PR #234, opened 2024-07-23
feat: add a debug logger for a better develop experience

YCK1130:
Because

- We need a simple interface to log structs and maps prettily.
- Current jsonl logger format is difficult to read while developing

This commit

- Implement a logger that can auto-adjust to and expand any type of object

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/cohere-ai/cohere-go/v2 v2.8.5
 	github.com/emersion/go-imap/v2 v2.0.0-beta.3
 	github.com/emersion/go-message v0.18.1
+	github.com/fatih/color v1.17.0
 	github.com/fogleman/gg v1.3.0
 	github.com/frankban/quicktest v1.14.6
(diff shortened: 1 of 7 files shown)


chuang8511:
hi @YCK1130 
Thanks for doing this!

Want to check have you checked how to integrate this better logger into pipeline-backend?
If this logger can be integrated to our current logger, it will be super helpful for all developers.

image

YCK1130:
Hi @chuang8511, I have tried to integrate my logging functions into Zap. However, the Zap logger only accepts strings and its predefined fields, which contradicts the intention to automatically expand the data structure. Therefore, the only solution is to wrap the Zap logger in our own logger and process the input into a string before logging. I'm not sure if this meets the requirements.

chuang8511:
> Therefore, the only solution is to wrap the Zap logger in our own logger and process the input into a string before logging.
Seems ok for me. 
I think the main point is that we can make the log clearer and complete for the developers.

YCK1130:
In this case, we may need to take out the logger to a new repository to avoid the import cycle.

PR #235, opened 2024-07-23
feat: add token count for each chunk

chuang8511:
Because

- to build RAG, it is crucial to know the token count for each chunk
- it is good to analyse the raw text with the token count

This commit

- add token count for each chunk
- differentiate two token count between sum up token counts in chunks and raw text token count


operator/json/v0/README.mdx
```diff
@@ -82,6 +82,7 @@ Process JSON through a `jq` command
 | Task ID (required) | `task` | string | `TASK_JQ` |
 | JSON value | `json-value` | any | JSON entity to be processed by the filter. It can be any valid JSON datatype (e.g. number, string, hash, array). |
 | Filter (required) | `jq-filter` | string | Filter, in `jq` syntax, that will be applied to the JSON input |
+| (DEPRECATED) JSON string | `json-string` | string | (DEPRECATED, use 'JSON value' instead) String with the JSON value to be processed. This field allows templated inputs, but the data might require preprocessing (marshalling). This field will be used in absence of 'JSON value' for backwards compatibility reasons. |
```
donch1989, line 85: It seems like you're using an outdated version of compogen. The deprecated field shouldn't be displayed here.

donch1989 (approved):

PR #238, opened 2024-07-24
feat: add video operator to fulfil unstructured data process

chuang8511:
Because

- we want to make vdp process all types of unstructured data

This commit

- add video operator to subsample video to video or images for different use cases


operator/video/v0/video_operation.go
```diff
@@ -0,0 +1,185 @@
...
+	videoBytes, err := base64.StdEncoding.DecodeString(base.TrimBase64Mime(base64Video))
+
+	if err != nil {
+		return nil, fmt.Errorf("error in decoding for inner: %s", err)
+	}
+
+	videoPrefix := strings.Split(base64Video, ",")[0]
+
+	// TODO: chuang8511 map the file extension to the correct format
+	// using mp4 first because it has smaller size.
```
donch1989, line 64: The file extension name doesn't actually determine the size. MP4 is just a container, the video codec, such as H.264 or H.265, will affect the file size.

operator/video/v0/video_operation.go
```diff
@@ -0,0 +1,185 @@
...
+	tempFileIn := "temp.mp4"
+
+	err = os.WriteFile(tempFileIn, videoBytes, 0644)
```
donch1989, line 67: We should avoid using constant names for storing video files, as it could cause issues if we receive two requests simultaneously.
Maybe we can try using https://pkg.go.dev/os#CreateTemp to create a tempfile.

donch1989 (changes requested):

operator/video/v0/video_operation.go
chuang8511: fix.
For frame one, I do not use CreateTemp because 1 run will generate the multiple frames.
It means we cannot know how many frames it will generate in advance.
To fix it, I just set uuid as prefix to make it avoid duplicate.

operator/video/v0/video_operation.go
```diff
@@ -0,0 +1,200 @@
...
+		data, err := os.ReadFile(file)
+		if err != nil {
+			return nil, fmt.Errorf("error reading file %s: %v", file, err)
+		}
+
+		encoded := base64.StdEncoding.EncodeToString(data)
+
+		frames = append(frames, Frame(jpegPrefix+encoded))
+
+		err = os.Remove(file)
```
donch1989, line 178: we should use `defer` to make sure the files will be removed.

operator/video/v0/video_operation.go
```diff
@@ -0,0 +1,195 @@
...
+	files, err := filepath.Glob(random + "_frame_*.jpeg")
+	if err != nil {
+		return nil, fmt.Errorf("error listing frames: %s", err)
+	}
+
+	sort.Strings(files)
+	jpegPrefix := "data:image/jpeg;base64,"
+	var frames []Frame
+	for _, file := range files {
+		defer os.Remove(file)
```
donch1989, line 168: You should place the `defer` statement before the loop. Otherwise, if the loop returns early, some images won't be deleted.

chuang8511: modified!

donch1989 (approved):

PR #240, opened 2024-07-25
fix: change md format to html tag for correct frontend link

YCK1130:
Because

- Most frontend links are broken now, and some text is out of box.
- Change to HTML <a> tag can render a proper hyperlink

This commit

- Change markdown format link in .json file to HTML <a> tag

### Important Note:
Most links that refer to our own documentation are dead. Maybe we should merge this PR after fixing it.


donch1989:
> Most links that refer to our own documentation are dead. Maybe we should merge this PR after fixing it.

Hi @YCK1130 , could you share which ones are dead?

PR #244, opened 2024-07-26
feat: add a hook to avoid we miss make document

chuang8511:
Because

- we sometimes forgot to make gen-doc to update document

This commit

- add pre-commit hook to avoid missing updating document


.hooks/chech_mdx_changes.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
donch1989, line 1: File name is wrong.

donch1989: And do we need this script?
I think we can just always run `make build-doc` and `make gen-doc` directly in the hooks.

chuang8511: [tested](https://github.com/instill-ai/component/pull/251/files)

chuang8511: can we merge this because it is not user-face PR.

jvallesm (approved):

PR #245, opened 2024-07-26
chore(video, audio): update definition for document automation

chuang8511:
Because

- we want to automate documentation script

This commit

- add the chinese title for the components


donch1989:
@chuang8511
We don't need to maintain multi-language documentation support in components. I'll close this PR.

PR #257, opened 2024-07-30
feat: make component ID accessible on IExecution

jvallesm:
Because

- `pipeline-backend` needs to pass and later access the ID of the component that's being executed (i.e. the component ID within the recipe) for the logging feature. This requires modifying the `store.CreateExecution` signature, so I'm taking the opportunity to clean up a bit.
- Every component repeats the same wrapping. This is an opportunity to remove duplication and also concerns from the individual components (they shouldn't care about the wrapping, only about the core business logic in each component).
- There's also repetition in how the base component is created in every component. This can be simplified by passing the base execution and just modifying / extending what's needed.

This commit

- Modifies the `CreateExecution` signature to return an interface, not a wrapper implementation.
- Wraps the execution in a single place (`store.CreateExecution`).
- Receives the component ID on execution creation and makes it accessible in the `IExecution` interface.


jvallesm:
@chuang8511 let me know what you think about the changes in the `base` and `store` packages. cc @donch1989

donch1989:
> @chuang8511 let me know what you think about the changes in the `base` and `store` packages. cc @donch1989

Yeah, the original wrapper is somewhat counter-intuitive and hard to understand. I'm looking forward to this change.

jvallesm:
## QA

Tested on https://github.com/instill-ai/pipeline-backend/pull/577#issuecomment-2258740713

store/store.go
```diff
@@ -156,12 +156,46 @@ func (s *Store) Import(comp base.IComponent) {
...
+	ComponentDefinitionID string
+
+	// SystemVariables contains information about the pipeline trigger in which
+	// the component is being executed.
+	SystemVariables map[string]any
+
+	// Setup may contain the configuration to connect to an external service.
+	Setup *structpb.Struct
+
+	// Task determines the task that the execution will carry out. It the input
```
tillknuesting, line 175: *It is the input

tillknuesting:
lgtm

store/store.go
```diff
@@ -156,12 +156,46 @@ func (s *Store) Import(comp base.IComponent) {
...
+		ComponentID:     p.ComponentID,
+		SystemVariables: p.SystemVariables,
+		Setup:           p.Setup,
+		Task:            p.Task,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("creating component execution: %w", err)
+	}
+
+	return &base.ExecutionWrapper{IExecution: x}, nil
```
donch1989, line 198: A clever change 🚀

donch1989 (approved):
LGTM

chuang8511:
LGTM! Just write my thoughts as a record.

In the long-term future, when we start to consider failover in the perspective of component, the execution seems to be instance. So, execution will need its ID as well. 
(Relationship between component & execution is component ||-|< execution)

I guess it is a long term future. 
When we decide to do failover system, we will need to reconsider how to deal with failover feature.
The current design seems to have the extensibility in this point of view.

note: not sure if failover is a good word.

jvallesm:
> execution will need its ID as well

The way we discussed it for the logging feature was that the pipeline trigger ID + the component ID form a unique key, which was enough for this feature. That might be an option for failover, assigning an Execution ID is also valid.

Thanks for the reviews 💞 I'll proceed with the merge

PR #258, opened 2024-07-30
chore: remove debug logger in jira

YCK1130:
Because

- We don't need debug logger in merged component

This commit

- remove the debug logger


donch1989:
Hi @YCK1130 
I tried to resolve the conflict, but it still needs some adjustment. Could you help resolve the conflict so we can merge it?

PR #265, opened 2024-08-01
feat(hubspot): add 4 tasks and modify Retrieve Association task and Get Thread task

AmeliaCelline:
Because

- need to implement more tasks for HubSpot to make the user story more complete.

This commit

- 4 additional tasks will be implemented

TODO:
- [x] Update Deal
- [x] Update Ticket
- [x] Get Owner (allow users to get HubSpot users information using either Owner ID or User ID)
- [x] Get All (allow users to get all the object IDs they want, such as contacts, companies, etc)
- [x] Fix Retrieve Association Task so that it can get more than 100+ object IDs (pagination)
- [x] Fix Get Thread task so that there is no limit in the number of messages it can get from a thread (pagination)


chuang8511 (approved):
lgtm

PR #272, opened 2024-08-05
fix: ignore bold case and add all line to result

chuang8511:
Because

- we want to support English when transform pdf to markdown
- we do not support bold in the current version

This commit

- remove loop to insert character but a line


chuang8511:
e2e test
image

[sample (1).pdf](https://github.com/user-attachments/files/16494428/sample.1.pdf)

PR #273, opened 2024-08-06
chore(web): refactor web to operator and add scrap webpage as a new task

chuang8511:
Because

- we are unifiying various file types, including web pages. Therefore, website component should be along with other similar operators, like text, image, audio, video operators.
- the current codebase scrapes webpage media, such as images, videos, and audios, using relative URLs, which results in non-functional media links.
- we want to enhance the ability of scraping a webpage for llm-ready data

This commit

- move website to operator and rename it to web
- add domain url in md converter to make image contains absolute url
- add new task to scrape a web page to make data llm-ready


operator/web/v0/config/definition.json
```diff
@@ -0,0 +1,18 @@
...
+    "TASK_SCRAPE_WEBPAGE"
+  ],
+  "documentationUrl": "https://www.instill.tech/docs/component/operator/web",
+  "icon": "assets/web.svg",
+  "id": "web",
+  "public": true,
+  "title": "Web",
+  "type": "COMPONENT_TYPE_OPERATOR",
+  "uid": "98909958-db7d-4dfe-9858-7761904be17e",
+  "version": "0.1.1",
```
jvallesm, line 14: This change affects the component definition list endpoint, used by e.g. the website. There's an index on the `pipeline-backend` database and, on initialisation, changes will be looked for so the index is updated. But the detected changes are version and visibility, we didn't think a component would change types, title or ID.

- Bumping the version will signal the initialiser to update the component info in the index.
- Regardless the component definition list, we should version our component definition changes.

```suggestion
  "version": "0.2.0",
```

Please QA how your changes affect https://www.instill.tech/component

chuang8511: Ah, I realised that we manage the component version by `ComponentDefinition`. right?
I guess it is the reason why we change json and all pipelines will adjust recipe.

So, we should check 
1. if all website component in the existing pipelines change to web and work correctly
2. [product website](https://www.instill.tech/components)
3. other I may miss.

chuang8511: Will refer to https://github.com/instill-ai/component/pull/249/files

donch1989: Actually, I forgot that we should update the version in that [PR](https://github.com/instill-ai/component/pull/249/files).

> if all website component in the existing pipelines change to web and work correctly
You need to write a migration script in pipeline-backend to change the component `type` in the existing recipe.

donch1989:
Hi @chuang8511 
Could you rebase the PR first? I can test it and add the migration script.

chuang8511:
migration
https://github.com/instill-ai/pipeline-backend/pull/587

PR #281, opened 2024-08-12
fix(artifact): add the description to remind users to add file extension

chuang8511:
Because

- the file name should include file extension as suffix

This commit

- add description for users


chuang8511:
I have tested use-case driven pipeline and it passed.
If there is further bug, I will create another PR again.

PR #282, opened 2024-08-12
feat(sql): add ssl/tls input as base64 encoded and move engine to setup

zidanehakim:
Because

- We need sql component to have a secure connection with ssl/tls
- Engine input within setup will be more robust as part of dsn connection

This commit

- Resolve those two things


zidanehakim:
Note:

- For SSL/TLS connections, users will input base64-encoded data for the (CA, cert, key) files, depending on whether they are using one-way or mutual TLS. This base64 data will be decoded, and temporary files will be created for referencing, as SQL Server only supports file paths for tls configuration.
- Since the SQL Server driver "github.com/denisenkom/go-mssqldb" only accepts a single certificate parameter (without separate cert and key), mutual TLS is not feasible with this driver (not sure whether just this driver or sql server itself). Thus, even if the user selects mutual TLS from the oneOf dropdown, the design will still use one-way TLS.

Important:

- The oneOf design, similar to the generic REST API component, allows users to choose their TLS type. However, due to a limitation or probably bug, the input cannot be referenced directly, so users must paste the base64-encoded string manually.

chuang8511 (dismissed):
lgtm.

chuang8511:
@zidanehakim Please rebase it! Thank you!

chuang8511:
@AmeliaCelline Sorry, I thought it could be fine to merge [this](https://github.com/instill-ai/component/pull/265). but it seems there is the error. Could you check it today?

zidanehakim:
@chuang8511 ready to be merged

PR #283, opened 2024-08-12
feat(groq, fireworksai): take out the unsupported models from instill credit

chuang8511:
Because

- the pricing info is not clear yet

This commit

- take out models that cannot support by instill credit now


chuang8511:
Please refer to google sheet, named AI models Pricing, attached in Linear.

PR #284, opened 2024-08-12
chore(artifact): update task name

chuang8511:
Because

- the name is confusing users

This commit

- change the name


chuang8511:
note
I do not do migration because we have not released to prod.

PR #285, opened 2024-08-13
fix(artifact): log the error that should not raise

chuang8511:
Because

- we do not have to raise error when the users plan to upload a file to a catalog

This commit

- log error but not raise error


chuang8511:
Will be done by another PR

chuang8511:
will be done by another PR

PR #290, opened 2024-08-19
feat(artifact): add retrieve messages and write message tasks

chuang8511:
Because

- we want catalog to able to have the memory system to achieve AI assistant

This commit

- implement the tasks about the messages


chuang8511:
There are 2 things that we have to check and do
1. e2e test
2. test code

data/artifact/v0/config/tasks.json
```diff
@@ -1029,36 +1096,338 @@
...
+              "update-time"
+            ],
+            "title": "Message",
+            "type": "object"
+          },
+          "instillFormat": "array:object",
+          "title": "Messages",
+          "type": "array"
+        },
+        "next-page-token": {
```
donch1989, line 1250: We don't have page-token in input, there is no need to return next-page-token.

data/artifact/v0/config/tasks.json
```diff
@@ -1029,36 +1096,338 @@
...
+          "items": {
+            "properties": {
+              "message-uid": {
+                "description": "The unique identifier of the message",
+                "instillFormat": "string",
+                "instillUIOrder": 0,
+                "title": "Message ID",
+                "type": "string"
+              },
+              "catalog-id": {
```
donch1989, line 1183: I think all messages in the array should share the same `catalog-id` and `conversation-id`, also matching the user's input. There's no need to return them in the output again.

donch1989 (changes requested):

chuang8511:
Convert it to draft because we will modify it after standardised AI task.

data/artifact/v0/chat_operation.go
```diff
@@ -0,0 +1,239 @@
...
+
+	artifactClient, connection := e.client, e.connection
+	defer connection.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+
+	ctx = metadata.NewOutgoingContext(ctx, getRequestMetadata(e.SystemVariables))
+
+	_, err = artifactClient.CreateConversation(ctx, &artifactPB.CreateConversationRequest{
```
chuang8511, line 199: please use ListConversations.
If there is no Conversation, call CreateConversation.

PR #291, opened 2024-08-20
chore(document): update doc and key to make vdp consistent

chuang8511:
Because

- we want to make doc clear and vdp consistent

This commit

- update the doc and field in document operator


donch1989 (approved):
LGTM, please send a pipeline-backend PR to migrate it.

PR #295, opened 2024-08-22
feat(openai): use go-openai client

donch1989:
Because

- Originally, we used `httpclient` to connect directly to OpenAI. Since we are introducing the streaming feature, using the `go-openai` SDK will make the codebase easier to maintain.

This commit

- Switches to using the `go-openai` client.


ai/groq/v0/groq_client_interface_mock_test.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by http://github.com/gojuno/minimock (v3.3.13). DO NOT EDIT.
+// Code generated by http://github.com/gojuno/minimock (v3.3.11). DO NOT EDIT.
```
jvallesm, line 1: FYI @donch1989 I created a [`make gen-mock`](https://github.com/instill-ai/component/blob/main/Makefile#L8) command to keep the mocks at the latest `minimock` version.

base/io/struct.go
jvallesm, line 1: ❓ Isn't this file publicly available so we can import it instead of copying it? Or are we extending it with custom code? In that case, can we highlight the methods we added / modified?

ai/openai/v0/client.go
```diff
@@ -1,48 +1,21 @@
...
-func (e errBody) Message() string {
-	return e.Error.Message
-}
-
-// getBasePath returns OpenAI's API URL. This configuration param allows us to
-// override the API the connector will point to. It isn't meant to be exposed
-// to users. Rather, it can serve to test the logic against a fake server.
-// TODO instead of having the API value hardcoded in the codebase, it should be
-// read from a setup file or environment variable.
-func getBasePath(setup *structpb.Struct) string {
```
jvallesm, line 40: This is still relevant for the tests (when introduced), we can use the [`NewClientWithConfig`](https://github.com/sashabaranov/go-openai/blob/master/client.go#L55) method to test the calls against a test server

base/io/struct.go
donch1989: It was a mistake; I accidentally committed some incorrect files in this PR.

ai/openai/v0/client.go
donch1989: It turns out that `go-openai` has limited support for optional parameters, so I’m going to revert this PR.

PR #298, opened 2024-08-23
fix(hubspot): fix test code

chuang8511:
Because

- we change the interface, which makes test code failed

This commit

- fix test code


chuang8511:
cc @AmeliaCelline

PR #299, opened 2024-08-23
feat: add milvus component

zidanehakim:
Because

- We want to integrate VDP with milvus vector db
- We want more vector db component choice

This commit

- Add milvus component along with the basic tasks (vector search, upsert, ..)
- Use account (username and password) for self hosted milvus (they dont provide api key auth yet)


chuang8511 (commented):
I missed the file name. Please change the file name

chuang8511:
But, why do you convert to API? I saw Milvus Golang SDK are official one?

zidanehakim:
@chuang8511 Milvus golang SDK in my opinion isn't convenient to use as it uses grpc in the process.

1. I need to convert to possible data types, for example if user inputs [0.1,0.2] it could be either []float32 or []float64 depending on the collection configuration
2. For some reason golang sdk needs us to load the collection first before having to vector search the collection. Same goes for delete collection, we first need to release collection.
3. It's isnt actually the official golang sdk but rather a community developed and milvus put it as a reference on their website

By these tradeoffs, i've decided to change into rest API instead

GeorgeWilliamStrong (commented):
From documentation perspective, LGTM!

PR #305, opened 2024-08-27
feat: make the API key be optional for Instill-Credit-supported component

donch1989:
Because

- In the new low-code editor, users need to type the recipe manually. Asking users to write `${secret.INSTILL_SECRET}` is not user-friendly. We decided to make these fields optional; if the key is empty, the component will use built-in keys.

This commit

- Makes the API key optional for Instill-Credit-supported components.

### DIFF
--- a/ai/anthropic/v0/README.mdx
+++ b/ai/anthropic/v0/README.mdx
@@ -30,7 +30,7 @@ The component configuration is defined and maintained [here](https://github.com/
 
 | Field | Field ID | Type | Note |
 | :--- | :--- | :--- | :--- |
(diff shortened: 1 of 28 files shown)


jvallesm (dismissed):
Looks good 👍

jvallesm (commented):
⚠️ Don't forget to update the tests, most of them use empty credentials and now the component interprets it should resolve them to a non-existing injected credential. I think using `WithInstillCredential` in the test is the quickest way to fix them. Perhaps there's a case where we want to verify error on wrong resolution, I don't remember.

PR #307, opened 2024-08-28
fix(text): fix chunk position bugs

chuang8511:
Because

- some chunk will be skipped, so the output chunk count will be smaller than the actual chunks from split function

This commit

- fix the bug to get the idx from output chunk


donch1989 (approved):
LGTM

PR #336, opened 2024-09-12
chore(instillartifact): update component id and package name

chuang8511:
Because

- we want to align the component id and package name

This commit

- update component id and package name for artifact


jvallesm (approved):

chuang8511:
Let's merge this when the big data London ends.

PR #362, opened 2024-09-23
feat(universalai): add instill credit function for universal ai component

chuang8511:
Because

- we will support universal ai to execute vendors' ai models

This commit

- import the vendors' key to component and execution


jvallesm:
Ported to https://github.com/instill-ai/pipeline-backend/pull/680

PR #366, opened 2024-09-24
chore(instillapp): init instill app

chuang8511:
Because

- we will have instill app

This commit

- init instill app


jvallesm:
Ported to https://github.com/instill-ai/pipeline-backend/pull/681

PR #368, opened 2024-09-24
feat(text): add client sample code

chuang8511:
Because

- we want developers to be able to see the result without login into Instill Cloud

This commit

- expose text component function
- add the example section for golang client


jvallesm:
Ported to https://github.com/instill-ai/pipeline-backend/pull/682

PR #374, opened 2024-09-26
feat(document): repair pdf with libreoffic

chuang8511:
Because

- The file could be malformed

This commit

- repair pdf with libreoffice without using function in pdfplumber because of licence issue

### DIFF
--- a/operator/document/v0/execution/pdf_checker.py
+++ b/operator/document/v0/execution/pdf_checker.py
@@ -3,7 +3,9 @@ import json
 import base64
 import sys
 
-# TODO: Deal with the import error when running the code in the docker container
+# TODO chuang8511:
+# Deal with the import error when running the code in the docker container.
+# Now, we combine all python code into one file to avoid the import error.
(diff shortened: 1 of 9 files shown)


operator/document/v0/execution/pdf_checker.py
```diff
@@ -0,0 +1,21 @@
+from io import BytesIO
+import json
+import base64
+import sys
+
+# TODO: Deal with the import error when running the code in the docker container
```
donch1989, line 6: @chuang8511 
I think this PR is not ready, right?

chuang8511: > # TODO: Deal with the import error when running the code in the docker container

I will deal with this part with other PR.
But, this PR is not ready because there is a bug in the container for prod. I may need to modify this PR. So, I draft it first.

jvallesm:
Ported to https://github.com/instill-ai/pipeline-backend/pull/683

PR #381, opened 2024-09-28
fix(docs): fix tasks.json inconsistent issue

pinglin:
Because

- there are many wording inconsistency issue in the `tasks.json` files.
- duplicated `instillUIOrder` value causes the generated `README.mdx` inconsistent

This commit

- fix the issues


ai/anthropic/v0/config/tasks.json
```diff
@@ -227,7 +227,7 @@
             "reference",
             "template"
           ],
-          "title": "System message",
+          "title": "System Message",
```
jvallesm, line 230: nit: we should either document that titles **must** be title-cased in the contribution guidelines, or use the `titleCaseWithArticles` function [in the template](https://github.com/instill-ai/component/blob/main/tools/compogen/pkg/gen/readme.go#L164) to unify the titles across the documents. If we don't care particularly about the JSON definition, I'd go with the second option.

jvallesm (approved):
Looks good 👍 Thanks for spotting this!

